PR #5, opened 2022-12-14
Quota, Pricing and Sku information providers

giventocode:
This PR introduces the following changes (interrelated to the use Terra as Quota information provider):

1.  New interfaces that abstract key management plane functionality:

  - `IBatchQuotaProvider` allows batch quota providers other than ARM.
     - Includes an implementation for ARM, `ArmBatchQuotaProvider`
  - `IBatchSkuInformationProvider` abstracts how pricing and sizing information for supported Batch VMs is obtained.
      - `PriceApiBatchSkuInformationProvider` a new approach that uses local list of skus supported by Batch, with sizing and core information (`BatchSupportedVmSizeInformation.json`), and the retail pricing API to create a list of skus supported by Batch with compute and pricing information.
      - Changed implementation to use the new approach. 

2. Retail pricing API client, with caching support (`PriceApiClient`).

3. Batch and CosmosDB clients accept shared key authentication and MI. 
  - Created strongly typed config options for CosmosDB and Batch
    -  Feedback on where I need to change config in the deployer is welcome.  
  - If keys are provided, key authentication is used; otherwise MI is tried.
   
4. Addtional refactoring:
   - `BatchScheduler` uses the `IBatchQuotaProvider` and `IBatchSkuInformationProvider` in its flow to check quota and the minimum vm size for a given task.
   - Removed direct instantiation of loggers, cache, IRespository, and other dependencies in the start-up code.
   - Implemented strongly typed configuration options using the `IOptions` pattern.


src/TesApi.Web/DoOnceAtStartUpService.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        /// <summary>
+        /// Stops service.
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            logger.LogInformation("Service stopped");
```
MattMcL4475, line 69: Does this service ever stop?

src/TesApi.Web/DoOnceAtStartUpService.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            using (logger.BeginScope("Executing Start Up tasks"))
+            {
+                try
+                {
+                    logger.LogInformation("Executing Configuration Utils Setup");
+                    await ExecuteConfigurationUtilsSetupAsync();
```
MattMcL4475, line 41: What's the advantage of using a hosted service here, instead of just Task.Run() ?

MattMcL4475 (approved):

src/TesApi.Web/Management/PriceApiBatchSkuInformationProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+            return await appCache.GetOrAddAsync<List<VirtualMachineInformation>>(region, async () => await GetVmSizesAndPricesAsync(region));
+
+        }
+
+        private async Task<List<VirtualMachineInformation>> GetVmSizesAndPricesAsyncImpl(string region)
+        {
+            logger.LogInformation("Getting VM sizes and price information for region:{0}", region);
+
+            var localVmSizeInfoForBatchSupportedSkus = await GetLocalVmSizeInformationForBatchSupportedSkusAsync();
+            var pricingItems = await priceApiClient.GetAllPricingInformationForNonWindowsAndNonSpotVmsAsync(region).ToListAsync();
```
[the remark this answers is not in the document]
giventocode: yep, I will add this in my next pr, as I have a solution for this and the price.

src/TesApi.Web/DoOnceAtStartUpService.cs
giventocode: There is no running process, it is a singleton that after the start method is called, nothing remains running. This was a way to get something that gets wired up using DI, and avoids blocking the configuration startup process.

PR #18, opened 2022-12-28
Add an option to pass through a custom connection string to TES

jsaun:
Resolves https://github.com/microsoft/CromwellOnAzure/issues/457, TES already supports auth with Azure without managed identity with the AzureServicesAuthConnectionString environment variable, this adds the option to pass that through from the deployer. Tested by passing an app key from my aad account of the form RunAs=App;AppId={AppId};TenantId={TenantId};AppKey={ClientSecret} instead of the managed app id. https://learn.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication?view=azure-dotnet#use-a-shared-secret-credential-to-sign-into-azure-ad


MattMcL4475:
do we still need this?  is this still used for anything?

PR #20, opened 2022-12-28
Implement support for TES to use PostgreSql 

ashanhol:
-  This PR enables TES to use Azure Database for Postgres instead of Cosmos DB.
-  Removes all Cosmos DB support
-  Uses latest IOptions based configuration

CoA PR: https://github.com/microsoft/CromwellOnAzure/pull/567


src/deploy-tes-on-azure/Deployer.cs
```diff
@@ -1306,6 +1245,63 @@ await postgreSqlDnsZone
...
+                    var adminUser = configuration.PostgreSqlAdministratorLogin;
+
+                    if (configuration.UsePostgreSqlSingleServer)
+                    {
+                        adminUser = $"{configuration.PostgreSqlAdministratorLogin}@{configuration.PostgreSqlServerName}";
+                    }
+
+                    var commands = new List<string[]> {
+                        new string[] { "bash", "-lic", $"echo {configuration.PostgreSqlServerName}.postgres.database.azure.com:5432:{configuration.PostgreSqlTesDatabaseName}:{adminUser}:{configuration.PostgreSqlAdministratorPassword} >> ~/.pgpass" },
+                        new string[] { "chmod", "0600", "/home/tes/.pgpass" },
```
BMurri, line 1298: Does `~/.pgpass` not work here? While home directories are nearly always in a known subdirectory of the `/home` directory, that's not always the case, and if something really weird happens this won't work as expected.

BMurri (commented):
There's a couple other questions that I have that are related to behavior during upgrades, but there are other related issues in the existing code as well, so I'll address them separately

src/Tes/Repository/TesTaskPostgreSqlRepository.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        /// <summary>
+        /// Identical to GetItemsAsync, paging is not supported. All items are returned
+        /// </summary>
+        /// <param name="predicate">Predicate to query the TesTasks</param>
+        /// <param name="pageSize">Ignored and has no effect</param>
+        /// <param name="continuationToken">Ignored and has no effect</param>
+        /// <returns></returns>
+        public async Task<(string, IEnumerable<TesTask>)> GetItemsAsync(Expression<Func<TesTask, bool>> predicate, int pageSize, string continuationToken)
```
giventocode, line 162: I think is cleaner to make a concrete return type with these two values - a record. Alternatively, you could rename the method GetAllItemsAsync, and just return IEnumerable, and implement pagination later, as another method.

src/Tes/Utilities/ConnectionStringUtility.cs
```diff
@@ -0,0 +1,57 @@
...
+        public string GetPostgresConnectionString(
+            string postgreSqlServerName, 
+            string postgreSqlTesDatabaseName, 
+            string postgreSqlTesDatabasePort, 
+            string postgreSqlTesUserLogin, 
+            string postgreSqlTesUserPassword,
+            string postgreSqlServerNameSuffix = ".postgres.database.azure.com",
+            string sslMode = "VerifyFull")
+        {
+            if (string.IsNullOrEmpty(postgreSqlServerName))
```
giventocode, line 17: nit: we should agree on a consistent approach to validate the args - I've been using the newer approach of ArgumentNullException.ThrowIfNull.  Note: I don't have a strong preference, and I'd be okay to change my approach if that will drive consistency.

src/Tes/Utilities/ConnectionStringUtility.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using System.Text;
+
+namespace Tes.Utilities
+{
+    public class ConnectionStringUtility
+    {
+        public string GetPostgresConnectionString(
```
giventocode, line 8: I think passing the options object rather than the individual values is a more maintainable approach. PostgreSqlOptions can be added/removed/changed without changing the signature of the method.

giventocode (changes requested):
LGTM! Left a few comments.

src/Tes/Repository/TesTaskPostgreSqlRepository.cs
MattMcL4475: This method signature comes from the original interface for Cosmos DB; will create a separate issue to harmonize this once Cosmos DB completely removed

src/Tes/Utilities/ConnectionStringUtility.cs
MattMcL4475: Done

src/Tes/Repository/TesTaskPostgreSqlRepository.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        /// <summary>
+        /// Base class searches within model, this method searches within the JSON
+        /// </summary>
+        /// <param name="tesTask"></param>
+        /// <returns></returns>
+        public async Task<TesTask> UpdateItemAsync(TesTask tesTask)
+        {
+            using var dbContext = createDbContext();
+            var item = await dbContext.TesTasks.FirstOrDefaultAsync(t => t.Json.Id == tesTask.Id);
```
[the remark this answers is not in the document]
MattMcL4475: The `select` here (await... FirstOrDefaultAsync()) is completed in a single transaction, and the update in a single transaction.  No TCP connection is held open between the `FirstOrDefaultAsync()` and `SaveChangesAsync()`.  The select is so that EF can track changes and then generate update SQL that only modifies the changed properties.   PostgreSQL takes a row lock during the update.

src/deploy-tes-on-azure/Deployer.cs
MattMcL4475: @jsaun any thoughts?

src/Tes/Repository/TesTaskPostgreSqlRepository.cs
```diff
@@ -0,0 +1,169 @@
...
+    {
+        private readonly Func<TesDbContext> createDbContext;
+
+        /// <summary>
+        /// Default constructor
+        /// </summary>
+        /// <param name="connectionString">The PostgreSql connection string</param>
+        public TesTaskPostgreSqlRepository(string connectionString)
+        {
+            createDbContext = () => { return new TesDbContext(connectionString); };
```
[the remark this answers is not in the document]
MattMcL4475: It has since been modified so can no longer be simplified

src/deploy-tes-on-azure/Deployer.cs
MattMcL4475: Resolved by @jsaun

src/TesApi.Web/AzureProxy.cs
```diff
@@ -144,27 +144,11 @@ public static async Task<string> GetAppInsightsInstrumentationKeyAsync(string ap
...
-            {
-                throw new Exception($"CosmosDB account '{cosmosDbAccountName} does not exist or the TES app service does not have Account Reader role on the account.");
-            }
-
-            var key = (await azureClient.WithSubscription(account.Manager.SubscriptionId).CosmosDBAccounts.ListKeysAsync(account.ResourceGroupName, account.Name)).PrimaryMasterKey;
-
-            return (account.DocumentEndpoint, key);
-        }
-
-        /// <inheritdoc/>
```
BMurri, line 167: Why re-add explicit comments on methods defined in the interface?

src/TesApi.Web/appsettings.json
```diff
@@ -14,13 +14,11 @@
     "SubscriptionId": "",
     "ResourceGroup": ""
   },
-  "CosmosDb": {
-    "CosmosDbKey": "",
-    "CosmosDbEndpoint": "",
-    "AccountName": ""
-  },
-  "RetryPolicyOptions": {
```
BMurri, line 22: Are we no longer using RetryPolicyOptions?

BMurri (approved):
lgtm. Just a couple clarifying questions

Looking forward to the corresponding deployer changes in CoA

giventocode (approved):

src/TesApi.Web/appsettings.json
BMurri: Actually, I discovered that this is misnamed. As written, it does absolutely nothing. To be effective, it needs to be named `RetryPolicy`, not `RetryPolicyOptions`

BMurri (approved):
Still approved. Consider my new comment.

src/TesApi.Web/appsettings.json
MattMcL4475: done

PR #23, opened 2023-01-06
Add optional ingress for TES with https and basic auth

jsaun:


MattMcL4475 (approved):
Looks great!  Does LetsEncryptEmail need to be set by the user when they run the deployer?  If so we should update the docs.

PR #27, opened 2023-01-11
Add support for a generic executions directory for TES

jsaun:


src/TesApi.Web/BatchScheduler.cs
```diff
@@ -38,6 +38,7 @@ public class BatchScheduler : IBatchScheduler
         private const int DefaultMemoryGb = 2;
         private const int DefaultDiskGb = 10;
         private const string CromwellPathPrefix = "/cromwell-executions/";
+        private const string ExecutionsPathPrefix = "/executions";
```
MattMcL4475, line 41: maybe this should end with "/" to be consistent with the prefix above it?  then would need to change the way it's used in GetBatchExecutionDirectoryPath too

MattMcL4475 (approved):

PR #32, opened 2023-01-12
feature: add tes compliance tests to tes api integration test

yuliadub:
I thought the current docker image used for the TES integration test is broken, as the image itself doesnt contain the tesapi.dll that it is using for entry point - this was an attempt to see if I can fix it. Docker file the test is currently using: https://github.com/microsoft/ga4gh-tes/blob/develop/src/TesApi.Web/Dockerfile_PackagePublishedFiles

when I try to build and run the image from that Dockerfile, it exists immediately since it cant find tesapi.dll which is the entry point. 

taking a look at other tests (cromwell), it does seem that they take a similar approach where the entry point defined in the Dockerfile is not included in the image itself, so likely I am missing the e2e flow of how tests are performed. What I have been able to gather is that ADO builds the files, then copies the binaries into the release, where then docker build is run in the context of the dir that has all the output files, therefore copying the binaries into the image. Docker build gets run as part of the pushing of the image to container repo. (big thanks to Jonathon for helping me with this!) 

looks like TES test may be failing due to an error in [nginx ](https://ms.portal.azure.com/#view/Microsoft_Azure_ContainerService/K8ResourceEditor.ReactView/aksClusterId/%2Fsubscriptions%2Ff483a450-5f19-4b20-9326-b5852bb89d83%2FresourceGroups%2Ftesinttest-3.3.23009.6364%2Fproviders%2FMicrosoft.ContainerService%2FmanagedClusters%2Ftesinttest-1aa3880288e/isAdd~/false/readonly~/true/title/Retriable%3A%20false%2C%20RetryAfter%3A%200s%2C%20HTTPStatusCode%3A%20400%2C%20RawError%3A%20%7B%0D%0A%20%20%22error%22%3A%20%7B%0D%0A%20%20%20%20%22code%22%3A%20%22InvalidDomainNameLabel%22%2C%0D%0A%20%20%20%20%22message%22%3A%20%22The%20domain%20name%20label%20tesinttest-33230096364-5c5257993a1238c2fdac4e.southcentralus.cloudapp.azure.com%20is%20invalid.%20It%20must%20conform%20to%20the%20following%20regular%20expression%3A%20%5E%5Ba-z%5D%5Ba-z0-9-%5D%7B1%2C61%7D%5Ba-z0-9%5D%24.%22%2C%0D%0A%20%20%20%20%22details%22%3A%20%5B%5D%0D%0A%20%20%7D%0D%0A%7D/resource~/%7B%22kind%22%3A%22Event%22%2C%22name%22%3A%22ingress-nginx-controller.173960d88c2ef88d%22%2C%22namespace%22%3A%22coa%22%7D )instead. Although unclear to me why its actually unhappy with that value.


yuliadub:
closing since the history on this is useless, going to open a new clean PR with just the needed changes

PR #42, opened 2023-01-19
Terra client API parses url correctly, retries improvement

### LINKED ISSUE #41: Connection with Terra LZ Quota API fails when TES stars up
At start up, when TES is configured to use Terra, the application fails to start. The issue occurred when TES attempts to create the allowed vm sizes config file at start up. 

Root cause:
 - The Terra API client parses the Terra LZ API url incorrectly.

Also identified:
 - The retry policy does not correctly recreate the httpRequest in each retry.

giventocode:
Closes #41

### DIFF
--- a/src/TesApi.Web/Management/Clients/HttpApiClient.cs
+++ b/src/TesApi.Web/Management/Clients/HttpApiClient.cs
@@ -23,6 +23,8 @@ namespace TesApi.Web.Management.Clients
         private readonly SHA256 sha256 = SHA256.Create();
         private readonly ILogger<HttpApiClient> logger;
         private readonly string tokenScope;
+        private readonly object lockObject = new object();
+        private AccessToken accessToken;
 
         /// <summary>
         /// Inner http client.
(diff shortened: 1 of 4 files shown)


src/TesApi.Web/Management/Clients/HttpApiClient.cs
```diff
@@ -187,17 +196,56 @@ private async Task AddAuthorizationHeaderToRequestAsync(HttpRequestMessage reque
             }
         }
 
+        private async Task<string> GetOrRefreshAccessToken()
+        {
+            // here we are synchronizing access to the access token only, there is a still
+            // a small chance the call to GetTokenAsync won't be synchronized, and it will get call more than once, concurrently
+            // however the impact should be marginal - the cached token will be updated more than once. 
+            lock (lockObject)
```
MattMcL4475, line 204: new SemaphoreSlim(1,1) works well here instead of lock since it has an async `WaitAsync()` method.  Can then add `.Release()` in a `Finally` block

https://learn.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim?view=net-7.0

src/TesApi.Web/Management/Clients/TerraLandingZoneApiClient.cs
```diff
@@ -11,7 +11,7 @@ namespace TesApi.Web.Management.Clients
     /// </summary>
     public class TerraLandingZoneApiClient : HttpApiClient
     {
-        private const string LandingZonesApiSegments = @"/api/landingzones/v1/azure";
+        private const string LandingZonesApiSegments = @"/api/landingzones/v1/azure/";
         private const string TokenScope = @"https://management.azure.com/.default";
```
MattMcL4475, line 15: this will eventually need to be config because when running in Azure China or Azure Government, it has a different domain for the management plane

MattMcL4475 (approved):
Added some minor suggestions, thank you!

src/TesApi.Web/Management/Clients/HttpApiClient.cs
giventocode: I considered but one concern was performance.  This method is a low-level call, in the http pipeline, and I wanted to avoid the performance impact of adding to the blocking time the I/O operation and the synchronization overhead of the semaphore, if possible.  In short, I wanted to synchronize r/w access to the property, which I can do with a simple lock.

PR #47, opened 2023-01-24
Terra WSM API Client, SAS Token operation

giventocode:
- New API client, for the Terra WSM `TerraWsmApiClient`
  - The client can be used to get the SAS tokens from Terra.  
- Refactored token locking code to use `SemaphoreSlim`, instead of `lock`.

### DIFF
--- a/src/TesApi.Tests/TerraApiStubData.cs
+++ b/src/TesApi.Tests/TerraApiStubData.cs
@@ -6,10 +6,14 @@ namespace TesApi.Tests;
 
 public class TerraApiStubData
 {
-    public string ApiHost => "https://landingzone.host";
+    public string LandingZoneApiHost => "https://landingzone.host";
+    public string WsmApiHost => "https://wsm.host";
+
     public string ResourceGroup => "mrg-terra-dev-previ-20191228";
     public Guid LandingZoneId { get; } = Guid.NewGuid();
     public Guid SubscriptionId { get; } = Guid.NewGuid();
+    public Guid WorkspaceId { get; } = Guid.NewGuid();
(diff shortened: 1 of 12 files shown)


src/TesApi.Tests/TerraApiStubData.cs
```diff
@@ -6,10 +6,14 @@ namespace TesApi.Tests;
 
 public class TerraApiStubData
 {
-    public string ApiHost => "https://landingzone.host";
+    public string LandingZoneApiHost => "https://landingzone.host";
+    public string WsmApiHost => "https://wsm.host";
+
     public string ResourceGroup => "mrg-terra-dev-previ-20191228";
```
MattMcL4475, line 12: is this supposed to be here?

src/TesApi.Tests/TerraApiStubData.cs
```diff
@@ -6,10 +6,14 @@ namespace TesApi.Tests;
...
     public Guid LandingZoneId { get; } = Guid.NewGuid();
     public Guid SubscriptionId { get; } = Guid.NewGuid();
+    public Guid WorkspaceId { get; } = Guid.NewGuid();
+    public Guid ContainerResourceId { get; } = Guid.NewGuid();
 
     public string BatchAccountId =>
```
MattMcL4475, line 18: is this supposed to be here?

MattMcL4475 (approved):

src/TesApi.Tests/TerraApiStubData.cs
giventocode: Yeah this is stub/fake data for testing and to validate parsing

PR #51, opened 2023-01-27
Terra Storage Provider

giventocode:
- This PR introduces the Terra storage provider `TerraStorageProvider`
 - New options in `TerraOptions` to set the workspace, storage account, container name and resource id from Terra.
 -  Additional refactoring:
     -  Added a base class for the storage providers to encapsulate common functionality amongst them: `StorageProvider` is now abstract. 
     - The existing storage provider was renamed to `DefaultStorageProvider`
     - Moved storage related classes to the `Storage` namespace/folder.
     - Noticed that `configurationUtils` was called at start up while setting up the DI. This is a duplicate operation that was impacting start up time - this called is via the `DoOnceAtStartUp` hosted service.

### DIFF
--- a/src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
+++ b/src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
@@ -12,19 +12,17 @@ using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Auth;
 using Microsoft.WindowsAzure.Storage.Blob;
 
-namespace TesApi.Web
+namespace TesApi.Web.Storage
 {
     /// <summary>
     /// Provides methods for blob storage access by using local path references in form of /storageaccount/container/blobpath
     /// </summary>
-    public class StorageAccessProvider : IStorageAccessProvider
+    public class DefaultStorageAccessProvider : StorageAccessProvider
     {
-        private readonly ILogger logger;
-        private readonly IAzureProxy azureProxy;
         private const string CromwellPathPrefix = "/cromwell-executions/";
         private const string BatchPathPrefix = "/executions/";
+        private static readonly TimeSpan SasTokenDuration = TimeSpan.FromDays(3); //TODO: refactor this to drive it from configuration. 
         private readonly string defaultStorageAccountName;
-        private static readonly TimeSpan sasTokenDuration = TimeSpan.FromDays(3);
         private readonly List<ExternalStorageContainerInfo> externalStorageContainers;
 
         /// <summary>
@@ -33,12 +31,10 @@ namespace TesApi.Web
         /// <param name="logger">Logger <see cref="ILogger"/></param>
         /// <param name="configuration">Configuration <see cref="IConfiguration"/></param>
         /// <param name="azureProxy">Azure proxy <see cref="IAzureProxy"/></param>
-        public StorageAccessProvider(ILogger<StorageAccessProvider> logger, IConfiguration configuration, IAzureProxy azureProxy)
+        public DefaultStorageAccessProvider(ILogger<DefaultStorageAccessProvider> logger, IConfiguration configuration, IAzureProxy azureProxy) : base(logger, azureProxy)
         {
-            this.logger = logger;
-            this.azureProxy = azureProxy;
-
-            this.defaultStorageAccountName = configuration["DefaultStorageAccountName"];    // This account contains the cromwell-executions container
+            //TODO: refactor to use the options pattern. 
+            defaultStorageAccountName = configuration["DefaultStorageAccountName"];    // This account contains the cromwell-executions container
             logger.LogInformation($"DefaultStorageAccountName: {defaultStorageAccountName}");
 
             externalStorageContainers = configuration["ExternalStorageContainers"]?.Split(new[] { ',', ';', '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
@@ -59,45 +55,9 @@ namespace TesApi.Web
         }
 
         /// <inheritdoc />
-        public async Task<string> DownloadBlobAsync(string blobRelativePath)
+        public override async Task<bool> IsPublicHttpUrl(string uriString)
         {
-            try
-            {
-                return await this.azureProxy.DownloadBlobAsync(new Uri(await MapLocalPathToSasUrlAsync(blobRelativePath)));
-            }
-            catch
-            {
-                return null;
-            }
-        }
-
-        /// <inheritdoc />
-        public async Task<bool> TryDownloadBlobAsync(string blobRelativePath, Action<string> action)
-        {
-            try
-            {
-                var content = await this.azureProxy.DownloadBlobAsync(new Uri(await MapLocalPathToSasUrlAsync(blobRelativePath)));
-                action?.Invoke(content);
-                return true;
-            }
-            catch
-            {
-                return false;
-            }
-        }
-
-        /// <inheritdoc />
-        public async Task UploadBlobAsync(string blobRelativePath, string content)
-            => await this.azureProxy.UploadBlobAsync(new Uri(await MapLocalPathToSasUrlAsync(blobRelativePath, true)), content);
-
-        /// <inheritdoc />
-        public async Task UploadBlobFromFileAsync(string blobRelativePath, string sourceLocalFilePath)
-            => await this.azureProxy.UploadBlobFromFileAsync(new Uri(await MapLocalPathToSasUrlAsync(blobRelativePath, true)), sourceLocalFilePath);
-
-        /// <inheritdoc />
-        public async Task<bool> IsPublicHttpUrl(string uriString)
-        {
-            var isHttpUrl = Uri.TryCreate(uriString, UriKind.Absolute, out var uri) && (uri.Scheme.Equals(Uri.UriSchemeHttp, StringComparison.OrdinalIgnoreCase) || uri.Scheme.Equals(Uri.UriSchemeHttps, StringComparison.OrdinalIgnoreCase));
+            var isHttpUrl = TryParseHttpUrlFromInput(uriString, out var uri);
 
             if (!isHttpUrl)
             {
@@ -126,7 +86,7 @@ namespace TesApi.Web
         }
 
         /// <inheritdoc />
-        public async Task<string> MapLocalPathToSasUrlAsync(string path, bool getContainerSas = false)
+        public override async Task<string> MapLocalPathToSasUrlAsync(string path, bool getContainerSas = false)
         {
             // TODO: Optional: If path is /container/... where container matches the name of the container in the default storage account, prepend the account name to the path.
             // This would allow the user to omit the account name for files stored in the default storage account
@@ -136,7 +96,7 @@ namespace TesApi.Web
             {
                 path = $"/{defaultStorageAccountName}{path}";
             }
-
+            //TODO: refactor this to throw an exception instead of logging and error and returning null.
             if (!StorageAccountUrlSegments.TryCreate(path, out var pathSegments))
             {
                 logger.LogError($"Could not parse path '{path}'.");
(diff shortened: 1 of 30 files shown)


src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+        public override Task<bool> IsPublicHttpUrlAsync(string uriString)
+        {
+            //TODO: check if this assumption is correct
+            //For Terra, if it is a Http/Https Url then it is public.
+            return Task.FromResult(TryParseHttpUrlFromInput(uriString, out _));
+
+        }
+
+        /// <inheritdoc />
+        public override async Task<string> MapLocalPathToSasUrlAsync(string path, bool getContainerSas = false)
```
giventocode, line 55: @BMurri and @MattMcL4475 can you provide feedback on this logic here? For Terra the expectation is that inputs will be called using the code below. Which means the storage account and container name will be always the two first segments of the path. 

```scala
 val blobSegment = ".blob.core.windows.net"
  def transformBlobString(s: String): String = if (s.contains(blobSegment)) {
    s.replaceFirst("https:/", "").replaceFirst(blobSegment, "")
  } else s
```

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            this.terraWsmApiClient = terraWsmApiClient;
+            ArgumentNullException.ThrowIfNull(terraOptions);
+
+            this.terraOptions = terraOptions.Value;
+
+        }
+
+        /// <inheritdoc />
+        public override Task<bool> IsPublicHttpUrlAsync(string uriString)
```
giventocode, line 46: @BMurri and @MattMcL4475, the assumption for Terra is that all Http Urls will be public endpoints. Is this the best way to implement the assumption - is there any other case we need to consider?

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
BMurri: A `TesInput` (and for that matter, a `TesOutput`) has a `Path` property that is always executor-local (and is always a file-spec). It also has a `Url` property which represents the (possible) source (for `TesInput`) and destination (for `TesOutput`), which is the property that needs to possibly be converted to a URI and possibly need to have a SAS appended.


Most of the time `MapLocalPathToSasUrlAsync()` operates on local paths. Occasionally, as a convenience to the user, it is used for "URL"s that are expressed as if they were a local path, but either are pointing to a `[default-storage]/container/blob-path` or `/storage-account/container/blob-path` path-spec-like expression.

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
BMurri: `IsPublicHttpUrlAsync` means that it can be reasonably assumed that the URL will not require any appended SAS (either because it already has one or because it points to an azure storage account and TES does not believe that it has access to request a SAS from that account.


If you have a URL for which you have a reasonable expectation to have to obtain a SAS token, then you should return `false` and expect to receive a callback to get a token.

src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
```diff
@@ -167,15 +126,15 @@ public async Task<string> MapLocalPathToSasUrlAsync(string path, bool getContain
                         var policy = new SharedAccessBlobPolicy()
                         {
                             Permissions = SharedAccessBlobPermissions.Add | SharedAccessBlobPermissions.Create | SharedAccessBlobPermissions.List | SharedAccessBlobPermissions.Read | SharedAccessBlobPermissions.Write,
-                            SharedAccessExpiryTime = DateTime.Now.Add(sasTokenDuration)
+                            SharedAccessExpiryTime = DateTime.Now.Add(SasTokenDuration)
```
MattMcL4475, line 129: Let's use DateTimeOffset.UtcNow here to stay consistent with UTC times throughout the code

https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.storage.blob.sharedaccessblobpolicy.sharedaccessexpirytime?view=azure-dotnet

src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
```diff
@@ -167,15 +126,15 @@ public async Task<string> MapLocalPathToSasUrlAsync(string path, bool getContain
...
                         };
 
                         var containerUri = new StorageAccountUrlSegments(storageAccountInfo.BlobEndpoint, pathSegments.ContainerName).ToUri();
                         resultPathSegments.SasToken = new CloudBlobContainer(containerUri, new StorageCredentials(storageAccountInfo.Name, accountKey)).GetSharedAccessSignature(policy, null, SharedAccessProtocol.HttpsOnly, null);
                     }
                     else
                     {
-                        var policy = new SharedAccessBlobPolicy() { Permissions = SharedAccessBlobPermissions.Read, SharedAccessExpiryTime = DateTime.Now.Add(sasTokenDuration) };
+                        var policy = new SharedAccessBlobPolicy() { Permissions = SharedAccessBlobPermissions.Read, SharedAccessExpiryTime = DateTime.Now.Add(SasTokenDuration) };
```
MattMcL4475, line 137: DateTimeOffset.UtcNow

src/TesApi.Web/Storage/StorageAccessProvider.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// <inheritdoc />
+    public async Task<string> DownloadBlobAsync(string blobRelativePath)
+    {
+        try
+        {
+            return await this.azureProxy.DownloadBlobAsync(new Uri(await MapLocalPathToSasUrlAsync(blobRelativePath)));
+        }
+        catch
+        {
+            return null;
```
MattMcL4475, line 51: TODO throw?

src/TesApi.Web/Storage/StorageAccessProvider.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        return Uri.TryCreate(input, UriKind.Absolute, out uri) && (uri.Scheme.Equals(Uri.UriSchemeHttp, StringComparison.OrdinalIgnoreCase) || uri.Scheme.Equals(Uri.UriSchemeHttps, StringComparison.OrdinalIgnoreCase));
+    }
+
+    /// <summary>
+    /// True if the path is the cromwell or executions folder
+    /// </summary>
+    /// <param name="path"></param>
+    /// <returns></returns>
+    protected bool IsItKnownExecutionFilePath(string path)
```
MattMcL4475, line 100: "IsKnownExecutionFilePath" makes more sense (can drop the word It)

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            if (IsItKnownExecutionFilePath(normalizedPath))
+            {
+                return await GetMappedSasUrlFromWsmAsync(normalizedPath);
+            }
+
+            if (!StorageAccountUrlSegments.TryCreate(normalizedPath, out var segments))
+            {
+                throw new Exception(
+                    "Invalid path provided. The path must be a valid blob storage url or a path with the following format: /accountName/container");
```
MattMcL4475, line 89: Can the code ever get here if the URL is a random Internet file?  I assume not...

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
```diff
@@ -0,0 +1,161 @@
...
+            return urlBuilder.ToString();
+        }
+
+        private async Task<string> GetMappedSasUrlFromWsmAsync(string blobName)
+        {
+            var tokenParams = CreateTokenParamsFromOptions(blobName, SasBlobPermissions);
+
+            var tokenInfo = await GetSasTokenFromWsmAsync(tokenParams);
+
+            logger.LogInformation($"Successfully obtained the Sas Url from Terra. Requested blobName:{blobName}. Wsm resource id:{terraOptions.WorkspaceStorageContainerResourceId}");
```
MattMcL4475, line 115: The blobName could contain customer information in the filename.  Usually it is something like a sample ID, but still would be better not to log it

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
```diff
@@ -0,0 +1,161 @@
...
+                Guid.Parse(terraOptions.WorkspaceId),
+                Guid.Parse(terraOptions.WorkspaceStorageContainerResourceId),
+                tokenParams);
+        }
+
+        private void CheckIfAccountAndContainerAreWorkspaceStorage(string accountName, string containerName)
+        {
+            if (!IsTerraWorkspaceStorageAccount(accountName))
+            {
+                throw new Exception($"The account name does not match. Expected:{accountName} and Provided{terraOptions.WorkspaceStorageAccountName}");
```
MattMcL4475, line 143: account name could contain customer information

MattMcL4475 (approved):
Minor comments; might want to remove account, container, blob names from the logs, and add more "todo" notes for throwing exceptions

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        private async Task<string> GetMappedSasContainerUrlFromWsmAsync(string pathToAppend)
+        {
+            //an empty blob name gets a container Sas token
+            var tokenInfo = await GetSasTokenFromWsmAsync(CreateTokenParamsFromOptions(blobName: string.Empty, SasContainerPermissions));
+
+            var urlBuilder = new UriBuilder(tokenInfo.Url);
+
+            urlBuilder.Path += pathToAppend;
```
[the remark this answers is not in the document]
giventocode: No, the path only has the path part of the URI - not the query string, which contains the query string.

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
giventocode: Okay. I see existing log entries that have this information. I will add an issue to look everywhere. I will remove this in my next PR for now.

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
```diff
@@ -0,0 +1,161 @@
...
+            return new SasTokenApiParameters(
+                terraOptions.SasAllowedIpRange,
+                terraOptions.SasTokenExpirationInSeconds,
+                sasPermissions, blobName);
+        }
+
+        private async Task<WsmSasTokenApiResponse> GetSasTokenFromWsmAsync(SasTokenApiParameters tokenParams)
+        {
+            logger.LogInformation(
+                $"Getting Sas Url from Terra. Requested blobName:{tokenParams.SasBlobName}. Wsm resource id:{terraOptions.WorkspaceStorageContainerResourceId}");
```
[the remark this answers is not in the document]
giventocode: Same as above.

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
giventocode: Makes sense. Will fix in my next PR.

src/TesApi.Web/Storage/TerraStorageAccessProvider.cs
giventocode: This code was updated.

src/TesApi.Web/Storage/StorageAccessProvider.cs
giventocode: Makes sense. Will fix in my next PR.

src/TesApi.Web/Storage/StorageAccessProvider.cs
giventocode: Yes, agree. I will add an issue for this as this was the existing flow.

src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
giventocode: Same as above, this is existing code.

src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
giventocode: Good point. I will add an issue for this, as the existing implementation does not use UTC, and I didn't want to change it as part of this.

Issue #43, opened 2023-01-19
AllocationFailed: The selected VM size 'Standard_D1_v2' cannot boot Hypervisor Generation '2'.


BMurri:
**Describe the bug**
In some azure locations (including the location we run our integration tests), we are seeing `vmsize`s that are not compatible with our default compute node images.

**Steps to Reproduce**
Create a PR.

**Expected behavior**
All eligible sizes should work with the configured image(s).

**Deployment details: (any information you can provide would be helpful):**
 - OS: n/a
 - Version: 3.3 & 4.0 (not yet released)
 - Any notable Azure customizations and/or configurations: none

**Example Error from Batch Account (Azure Portal)**

> Code: AllocationFailed Message:
> Desired number of low priority nodes could not be allocated Values:
> Provider Error Json - {
>   "error": {
>     "code": "BadRequest",
>     "message": "The selected VM size 'Standard_D1_v2' cannot boot Hypervisor Generation '2'. If this was a Create operation please check that the Hypervisor Generation of the Image matches the Hypervisor Generation of the selected VM Size. If this was an Update operation please select a Hypervisor Generation '2' VM Size. For more information, see [https://aka.ms/azuregen2vm](https://aka.ms/azuregen2vm)"
>   }
> }
> Provider Error Json Truncated - False

**Additional context**
There is a reason we changed the default image from a gen_1 to a gen_2, but I don't recall the exact specifics at the moment.

The public price API doesn't indicate with which generation each `vmsize` is compatible. The management-plane APIs we previously used didn't provide the same names (so we created an internal "allow list" we had to maintain for each release that could very well become out of date the day we release).

The only APIs that indicate which images are gen_1 or gen_2 and the only APIs that indicate which generations each `vmsize` is compatible are all in the management-plane.

Note that, while we don't expose the default image in either deployer, they can be changed by altering the configuration and running an update with the deployer.

Note also, that the not-yet fully implemented "host configuration" feature overrides/replaces the configured default image with the configured image from the host-configuration metadata (if supplied).

Some possible solutions without (re)adding more management-plane API calls:
| Whitelist | Single Gen | Description |
|:---------:|:----------:| ----------- |
| no | no | Change the current default image properties to two, one for each generation, when one image produces errors for the initial node, replace the pool with the another one using the other image. |
| no | yes | Essentially do nothing. |
| yes | no | Restore a white list encoding the generational compatibility, change the current default image properties to two, one for each generation. |
[...]

PR #52, opened 2023-01-27
Add program to generate list of batch supported skus

### LINKED ISSUE #34: Create console application to generate BatchSupportedVmSizeInformation.json and add to build
1.  Create a console application that uses the Batch API and add it to this repo that will generate this file:

BatchSupportedVmSizeInformation.json

1b.  Consult with @giventocode and @tonybendis on the existing implementation
2.  Run (1) during the build and update the file

Since the Batch API call is a management plane operation, in the future, this could potentially be done at runtime.

jsaun:
Fixes #34 and #43. Add command line program to generate BatchSupportedVmSizeInformation.json, adding vm generation and regions supported to the file also. Use the new fields to choose the correct vm image and filter to vm only available in current region.


jsaun:
Integration test currently failing due to a combination of #43 and new skus being added to BatchSupportedVmSizeInformation.json. It hits 3 skus that only support gen1 vm images and then fails, develop integration test is passing because it has fewer skus and finds a gen2 sku before erroring out.

BMurri (commented):
I like what I can see (github won't let me see src\TesApi.Web\BatchSupportedVmSizeInformation.json). I would suggest renaming "Utilities" to something like GenerateBatchVmSkus, unless you instead change the command-line to require a GenerateBatchVmSkus argument that prefixes the entire command-line (so we can expand it in the future).

I would also suggest that you rename the gen2 batchNodeInfo vars to match the new gen1 pattern

src/Utilities/GenerateBatchVmSkus.cs
```diff
@@ -0,0 +1,163 @@
...
+                            regionsForVm[vm] = new HashSet<string>() { region };
+                        }
+
+                        if (!sizeForVm.ContainsKey(vm))
+                        {
+                            if (sizes is null)
+                            {
+                                sizes = await subscription.GetVirtualMachineSizesAsync(region).ToListAsync();
+                            }
+                            sizeForVm[vm] = sizes.FirstOrDefault(vmsize => vmsize.Name.Equals(vm));
```
MattMcL4475, line 101: `SingleOrDefault` would be better here, so that it throws if more than 1 is found (which would tell us the data is dirty or we are not considering something)

src/Utilities/GenerateBatchVmSkus.cs
```diff
@@ -0,0 +1,163 @@
...
+                        }
+
+                        if (!skuForVm.ContainsKey(vm))
+                        {
+                            if (skus is null)
+                            {
+                                skus = await subscription.GetComputeResourceSkusAsync(region).ToListAsync();
+                            }
+                            skuForVm[vm] = skus.FirstOrDefault(sku => sku.Name == vm);
```
MattMcL4475, line 110: SingleOrDefault

src/Utilities/GenerateBatchVmSkus.cs
```diff
@@ -0,0 +1,163 @@
...
+            Console.WriteLine($"Superset supportedSkuCount:{batchSupportedVmSet.Count}");
+
+            var batchVmInfo = batchSupportedVmSet.Select((s) =>
+            {
+                var sizeInfo = sizeForVm[s];
+                var sku = skuForVm[s];
+                if (sizeInfo is not null)
+                {
+                    var generationList = new List<string>();
+                    var generation = sku?.Capabilities.Where(x => x.Name.Equals("HyperVGenerations")).FirstOrDefault()?.Value;
```
MattMcL4475, line 131: SingleOrDefault .  Also why would "sku" ever be null?

src/Utilities/GenerateBatchVmSkus.cs
```diff
@@ -0,0 +1,163 @@
...
+
+                    if (generation is not null)
+                    {
+                        generationList = generation.Split(",").ToList();
+                    }
+
+                    int.TryParse(sku?.Capabilities.Where(x => x.Name.Equals("vCPUsAvailable")).FirstOrDefault()?.Value, out var vCpusAvailable);
```
MattMcL4475, line 138: SingleOrDefault

src/Utilities/GenerateBatchVmSkus.cs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            var client = new ArmClient(new DefaultAzureCredential());
+            static double ConvertMiBToGiB(int value) => Math.Round(value / 1024.0, 2);
+            var subscription = client.GetSubscriptionResource(new ResourceIdentifier($"/subscriptions/{configuration.SubscriptionId}"));
+
+            var regionsForVm = new Dictionary<string, HashSet<string>>();
+            var sizeForVm = new Dictionary<string, VirtualMachineSize>();
+            var skuForVm = new Dictionary<string, ComputeResourceSku>();
+
+            foreach (var region in typeof(AzureLocation).GetTypeInfo().DeclaredProperties.Where(p => p.PropertyType == typeof(AzureLocation)).Select(p => p.Name).ToList())
```
MattMcL4475, line 76: I like getting the regions from here but unfortunately it's missing some it seems.  For example, "westus3".  We need to find a better source for regions, or maybe update the NUGET?

src/Utilities/GenerateBatchVmSkus.cs
```diff
@@ -0,0 +1,163 @@
...
+                    {
+                        generationList = generation.Split(",").ToList();
+                    }
+
+                    int.TryParse(sku?.Capabilities.Where(x => x.Name.Equals("vCPUsAvailable")).FirstOrDefault()?.Value, out var vCpusAvailable);
+
+                    return new VirtualMachineInformation()
+                    {
+                        MaxDataDiskCount = sizeInfo.MaxDataDiskCount,
+                        MemoryInGB = ConvertMiBToGiB(sizeInfo.MemoryInMB ?? 0),
```
MattMcL4475, line 143: should this be renamed to MemoryInGiB ?  Also wouldn't it be a bug if 0?

src/Utilities/GenerateBatchVmSkus.cs
```diff
@@ -0,0 +1,163 @@
...
+                        VCpusAvailable = vCpusAvailable,
+                        ResourceDiskSizeInGB = ConvertMiBToGiB(sizeInfo.ResourceDiskSizeInMB ?? 0),
```
MattMcL4475, line 145: ResourceDiskSizeInGiB ?  Also how could this ever be 0?

src/Utilities/GenerateBatchVmSkus.cs
```diff
@@ -0,0 +1,163 @@
...
+                        VmSize = sizeInfo.Name,
+                        VmFamily = sku?.Family,
```
MattMcL4475, line 147: Why would sku ever be null here?  Wouldn't that be a bug if no VmFamily?

src/Utilities/GenerateBatchVmSkus.cs
```diff
@@ -0,0 +1,163 @@
...
+            }
+            
+            var batchSupportedVmSet = regionsForVm.Keys.ToList();
+            Console.WriteLine($"Superset supportedSkuCount:{batchSupportedVmSet.Count}");
+
+            var batchVmInfo = batchSupportedVmSet.Select((s) =>
+            {
+                var sizeInfo = sizeForVm[s];
+                var sku = skuForVm[s];
+                if (sizeInfo is not null)
```
MattMcL4475, line 128: why would it ever be null?

MattMcL4475 (changes requested):
It seems we still need to add pricing to this too?

src/Utilities/GenerateBatchVmSkus.cs
MattMcL4475: az account list-locations -o table

MattMcL4475: you could check the az cli code to see which API it is using to get this list.  Notice it includes westus3

MattMcL4475 (approved):
- Make the minor changes
- Generate the initial list from a PAYG subscription
- Add the remaining work to a new issue

src/Utilities/GenerateBatchVmSkus.cs
jsaun: Looks like this did the trick without having to run on a pay-as-you-go sub: 
await subscription.GetLocationsAsync().Where(x => x.Metadata.RegionType == RegionType.Physical).ToListAsync();

src/Utilities/GenerateBatchVmSkus.cs
jsaun: It won't ever be 0, the null-coalesce is just there to cast int? to int before doing math.

src/Utilities/GenerateBatchVmSkus.cs
jsaun: It won't ever be 0, the null-coalesce is just there to cast int? to int before doing math.

MattMcL4475: In that case would recommend sizeInfo.MemoryInMB.Value , so the intention is clear

MattMcL4475: Also maybe MemoryInMB should be of type "int" if it can/should never be null

PR #57, opened 2023-02-01
Handle null inputs or outputs in a TES task

MattMcL4475:
Fixes the case when no inputs or outputs are specified like in the below JSON, which is accepted by TES but throws an exception while processing.

```
{
  "resources": {
    "preemptible": true,
    "backend_parameters": {
      "vm_size": "Standard_D2a_v4"
    }
  },
  "executors": [
    {
      "image": "ubuntu:22.04",
      "command": [ "echo hello > test.txt" ],
    }
  ]
}
```

### DIFF
--- a/src/TesApi.Web/BatchScheduler.cs
+++ b/src/TesApi.Web/BatchScheduler.cs
@@ -890,9 +890,14 @@ namespace TesApi.Web
             var downloadFilesScriptUrl = await this.storageAccessProvider.MapLocalPathToSasUrlAsync($"/{downloadFilesScriptPath}");
             await this.storageAccessProvider.UploadBlobAsync($"/{downloadFilesScriptPath}", downloadFilesScriptContent);
 
-            var filesToUpload = await Task.WhenAll(
-                task.Outputs?.Select(async f =>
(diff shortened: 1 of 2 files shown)


src/TesApi.Web/BatchScheduler.cs
```diff
@@ -823,25 +823,28 @@ private async Task<CloudTask> ConvertTesTaskToBatchTaskAsync(string taskId, TesT
...
+                .ToList() ?? new List<string>();
 
-            var inputFiles = task.Inputs.Distinct();
+            var inputFiles = task.Inputs?.Distinct().ToList() ?? new List<TesInput>();
 
             var drsInputFiles = inputFiles
                 .Where(f => f?.Url?.StartsWith("drs://", StringComparison.OrdinalIgnoreCase) == true)
                 .ToList();
 
-            foreach (var output in task.Outputs)
```
BMurri, line 840: You could avoid the additional if statement with "foreach (var output in task.Outputs ?? Array.Empty<TesOutput>())"

BMurri (approved):

src/TesApi.Web/BatchScheduler.cs
BMurri: Although you may need to cast it to IEnumerable<TesOutput>

BMurri: Enumerable.Empty\<TesOutput\>() may prevent needing to cast anything

PR #58, opened 2023-02-01
Remove task ID from executions path

MattMcL4475:
The task ID is created by TES and is not known _a priori_, so there is no way for the user to specify it.  Also it's unnecessary, since the task is already running in a unique path specified in the environment variable `$AZ_BATCH_TASK_WORKING_DIR`

This PR simply removes it from the path.  TES inputs and outputs shall now be specified like this:

`/executions/my_output_file.txt`

For a later discussion:

- There is no way the user would know that `/executions` is hard-coded in this implementation without reading the code or docs: https://github.com/microsoft/ga4gh-tes/blob/92208718b1d36d2eb2854f2ecd9892b3d9056a96/src/TesApi.Web/BatchScheduler.cs#L52
- Just using root seems better `/my_output_file.txt`, however there might be a permissions issue writing there
- Furthermore, it seems best to also support simply `my_output_file.txt` - however, [the TES spec says the path should be absolute](https://github.com/ga4gh/task-execution-schemas/blob/1df37d34242f74d3f03475c6b9de3324b8094054/openapi/task_execution_service.openapi.yaml#L537), when there is no slash, is it considered absolute still?
- There might be an unrelated bug here I noticed, since the first slash `/` is specified here:
https://github.com/microsoft/ga4gh-tes/blob/92208718b1d36d2eb2854f2ecd9892b3d9056a96/src/TesApi.Web/BatchScheduler.cs#L990
But not here:
https://github.com/microsoft/ga4gh-tes/blob/92208718b1d36d2eb2854f2ecd9892b3d9056a96/src/TesApi.Web/BatchScheduler.cs#L994
Or here:
https://github.com/microsoft/ga4gh-tes/blob/92208718b1d36d2eb2854f2ecd9892b3d9056a96/src/TesApi.Web/BatchScheduler.cs#L995

### DIFF
--- a/src/TesApi.Web/BatchScheduler.cs
+++ b/src/TesApi.Web/BatchScheduler.cs
@@ -821,7 +821,8 @@ namespace TesApi.Web
         private async Task<CloudTask> ConvertTesTaskToBatchTaskAsync(string taskId, TesTask task, bool poolHasContainerConfig)
         {
             var cromwellExecutionDirectoryPath = GetCromwellExecutionDirectoryPath(task);
-            var batchExecutionPathPrefix = cromwellExecutionDirectoryPath is null ? ExecutionsPathPrefix : CromwellPathPrefix;
+            bool isCromwell = cromwellExecutionDirectoryPath is not null;
+            var batchExecutionPathPrefix = isCromwell ? CromwellPathPrefix : ExecutionsPathPrefix;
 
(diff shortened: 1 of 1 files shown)


MattMcL4475:
Still has an issue that all the files will end up in the same container for all tasks; might be able to just change the upload path

BMurri (approved):

PR #77, opened 2023-02-09
Expose batch pool creation errors

BMurri:


src/TesApi.Web/BatchScheduler.cs
```diff
@@ -573,6 +526,70 @@ static bool IsPoolQuotaException(string code)
...
+                        logger.LogError(tesException, $"TES task: {tesTask.Id} TesException.Message: {tesException.Message}");
+                        break;
+
+                    case BatchClientException batchClientException:
+                        tesTask.State = TesState.SYSTEMERROREnum;
+                        tesTask.SetFailureReason("BatchClientException", string.Join(",", batchClientException.Data.Values), batchClientException.Message, batchClientException.StackTrace);
+                        logger.LogError(batchClientException, $"TES task: {tesTask.Id} BatchClientException.Message: {batchClientException.Message} {string.Join(",", batchClientException?.Data?.Values)}");
+                        break;
+
+                    case BatchException batchException when batchException.InnerException is Microsoft.Azure.Batch.Protocol.Models.BatchErrorException batchErrorException && IsJobQuotaException(batchErrorException.Body.Code):
```
giventocode, line 569: I believe these exceptions should be logged. It is hard to troubleshoot root causes if the root exception information is lost.

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -573,6 +526,70 @@ static bool IsPoolQuotaException(string code)
                     "PoolQuotaReached" => true,
                     _ => false,
                 };
+
+            void HandleException(Exception exception)
```
giventocode, line 530: This error handling seems complex to me. At the very least we should have a test that validates the desired behavior.

giventocode (changes requested):

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -573,6 +526,70 @@ static bool IsPoolQuotaException(string code)
...
+                        tesTask.SetFailureReason("InsufficientBatchQuota", azureBatchLowQuotaException.Message);
+                        logger.LogError(azureBatchLowQuotaException, $"TES task: {tesTask.Id} AzureBatchLowQuotaException.Message: {azureBatchLowQuotaException.Message}");
+                        break;
+
+                    case AzureBatchVirtualMachineAvailabilityException azureBatchVirtualMachineAvailabilityException:
+                        tesTask.State = TesState.SYSTEMERROREnum;
+                        tesTask.AddTesTaskLog(); // Adding new log here because this exception is thrown from GetVmSizeAsync() and AddTesTaskLog() above is called after that. This way each attempt will have its own log entry.
+                        tesTask.SetFailureReason("NoVmSizeAvailable", azureBatchVirtualMachineAvailabilityException.Message);
+                        logger.LogError(azureBatchVirtualMachineAvailabilityException, $"TES task: {tesTask.Id} AzureBatchVirtualMachineAvailabilityException.Message: {azureBatchVirtualMachineAvailabilityException.Message}");
```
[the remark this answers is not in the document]
BMurri: I haven't tested the logging solution we are using, but every other structured logging solution I've used in the past decade will include the chained inner exceptions in the log entries.

src/TesApi.Web/BatchScheduler.cs
BMurri: Same as above. I will, however, test to see what shows up in the logs.

src/TesApi.Web/BatchScheduler.cs
BMurri: I will add one

src/TesApi.Web/BatchScheduler.cs
BMurri: I double checked. In `Log Analytics Workspace`, under `LogManagement`, in addition to `AppTraces`, there's `AppExceptions`. Inside of that, the entire exception chain (following inner exceptions) including stack traces is present for every log where an exception was given to the API, as far as I can see. It's wonderful for root cause analysis.

BMurri:
@giventocode Tests added. Some minor edits elsewhere as well.

giventocode (approved):

PR #79, opened 2023-02-10
WSM and Terra Integration Test Fixes

### LINKED ISSUE #78: Batch pool creation fails on Terra
Batch pool creation fails with a bad request response 400, when calling Terra WSM to create a batch pool. 

The error is caused by several issues: 
 - Type mismatch between the interval for the autoscaling settings.
 - Incorrect casing in the `ApiDynamicVNetAssignmentScope` enumeration values.
 - Incorrect type in the `provision` property.

giventocode:
Closes #78 

- Fixes type and mapping issues to create a valid request to the WSM endpoint.
- Added exception logging to facilitate debugging
- Added required properties to the appsettings file.

### DIFF
--- a/src/TesApi.Web/Management/Clients/TerraWsmApiClient.cs
+++ b/src/TesApi.Web/Management/Clients/TerraWsmApiClient.cs
@@ -93,7 +93,7 @@ namespace TesApi.Web.Management.Clients
                     responseContent = await response.Content.ReadAsStringAsync();
                 }
 
-                Logger.LogError($"Failed to create a batch pool using wsm. Response content:{responseContent}", ex);
+                Logger.LogError(ex, $"Failed to create a batch pool using wsm. Response content:{responseContent}";
                 throw;
             }
         }
--- a/src/TesApi.Web/Management/Clients/TerraWsmApiClient.cs
+++ b/src/TesApi.Web/Management/Clients/TerraWsmApiClient.cs
@@ -93,7 +93,7 @@ namespace TesApi.Web.Management.Clients
                     responseContent = await response.Content.ReadAsStringAsync();
                 }
 
-                Logger.LogError(ex, $"Failed to create a batch pool using wsm. Response content:{responseContent}";
+                Logger.LogError(ex, $"Failed to create a batch pool using wsm. Response content:{responseContent}");
                 throw;
             }
         }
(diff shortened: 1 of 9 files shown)


src/TesApi.Web/appsettings.json
```diff
@@ -5,26 +5,33 @@
...
   "CosmosDb": {
-    "CosmosDbKey": "",
-    "CosmosDbEndpoint": "",
     "AccountName": ""
   },
   "RetryPolicyOptions": {
     "MaxRetryCount": 5,
     "ExponentialBackOffExponent": 2
-  },
-  "ContainerRegistry": {
```
MattMcL4475, line 26: Just checking you really meant to remove this?  if so, you can ignore this comment

MattMcL4475 (approved):

src/TesApi.Web/Management/Clients/TerraWsmApiClient.cs
```diff
@@ -68,13 +70,32 @@ public virtual async Task<ApiCreateBatchPoolResponse> CreateBatchPool(Guid works
...
+            }
+            catch (Exception ex)
+            {
+                var responseContent = "";
+                if (response is not null)
+                {
+                    responseContent = await response.Content.ReadAsStringAsync();
+                }
+
+                Logger.LogError($"Failed to create a batch pool using wsm. Response content:{responseContent}", ex);
```
BMurri, line 96: I thought that the exception goes before the message

src/TesApi.Web/Startup.cs
```diff
@@ -157,26 +160,36 @@ private void AddAppInsightsWithDefaultOrLookingUpTheConnectionString(IServiceCol
                 s.ConnectionString = $"InstrumentationKey={instrumentationKey}";
             });
         }
-
```
BMurri, line 160: There should be a blank line between methods/code blocks

BMurri (commented):
I'm concerned about the order of parameters to the LogError method. I also have the same question about configuration removals as Matt. Otherwise looks good to go.

BMurri (approved):

src/TesApi.Web/Management/Clients/TerraWsmApiClient.cs
giventocode: Thx. Will fix.

PR #88, opened 2023-02-14
SharedPool improvements

BMurri:
Pool names: prefix is larger and is now configurable (BatchPrefix).
RegEx for ResourceName is now documented.
AutoScale formula notes are less editorialized.
AutoScale formula is more readable.
Reduced number of exceptions when pool is manually deleted.
Late quota detection now covers jobs.
Pool names are prefixed with TES

CoA companion PR is microsoft/CromwellOnAzure#586


src/deploy-tes-on-azure/Configuration.cs
```diff
@@ -22,18 +21,11 @@ public class Configuration : UserAccessibleConfiguration
         public string PostgreSqlVersion { get; set; } = "11";
         public string DefaultPostgreSqlSubnetName { get; set; } = "sqlsubnet";
         public int PostgreSqlStorageSize { get; set; } = 128;  // GiB
-        public string Name { get; set; } = CreateNewName();
```
MattMcL4475, line 25: both deployers will need their AKS templates updated for these changes

BMurri: See microsoft/CromwellOnAzure#586

MattMcL4475 (approved):

PR #93, opened 2023-02-15
Match configuration and code

BMurri:


giventocode (approved):

MattMcL4475 (commented):
I think we are missing the updates to the helm chart in deploy-tes-on-azure here right?

BMurri:
@MattMcL4475 The helm chart didn't include `Gen2BatchNodeAgentSkuId`, so there was nothing to remove

Issue #100, opened 2023-02-17
Complete configuration implementation using IOptions


BMurri:
This is related to #53 and #63.

We are currently partially using IOptions for just three of its three main benefits (grouping related properties, improved type coercion, and ease of locating instances of use). We should not just extend that usage to as many of the remaining properties as is practical, but we should complete the implementation in order to take advantage of what is possibly its greatest benefit: incredible ease-of-use when implementing configuration monitoring (aka live configuration updates).

Currently our configuration consists of the following individual properties:

- Set in both deployment and appsettings.json

1. BatchAccount:AccountName
2. BatchNodeAgentSkuId
3. BatchPoolRotationForcedDays
4. DisableBatchNodesPublicIpAddress
5. DisableBatchScheduling
6. Gen1BatchImageOffer
7. Gen1BatchImagePublisher
8. Gen1BatchImageSku
9. Gen1BatchImageVersion
10. Gen2BatchImageOffer
11. Gen2BatchImagePublisher
12. Gen2BatchImageSku
13. Gen2BatchImageVersion
14. BranchPrefix
15. UseLegacyBatchImplementationWithAutopools
16. UsePreemptibleVmsOnly
17. TesPostgreSql:PostgreSqlServerName
18. TesPostgreSql:PostgreSqlServerNameSuffix
19. TesPostgreSql:PostgreSqlServerPort
20. TesPostgreSql:PostgreSqlServerSslMode
21. TesPostgreSql:PostgreSqlDatabaseName
22. TesPostgreSql:PostgreSqlDatabaseUserLogin
23. TesPostgreSql:PostgreSqlDatabaseUserPassword

- Set in deployment only

1. ApplicationInsightsAccountName
2. ASPNETCORE_URLS
3. AzureOfferDurableId
4. AzureServicesAuthConnectionString
5. BatchNodesSubnetId
6. BlobxferImageName
7. DefaultStorageAccountName
8. DockerInDockerImageName
9. ExternalStorageContainers
10. GlobalManagedIdentity
11. GlobalStartTaskPath
12. MarthaKeyVaultName
13. MarthaSecretName
14. MarthaUrl

- Set in appsettings.json only

1. AllowedHosts
2. Logging:LogLevel:Default
3. RetryPolicy:ExponentialBackOffExponent
4. RetryPolicy:MaxRetryCount

- Only configured in code

1. AllowedVmSizes

- (Mostly) used in code but never configured (note, this list might not be fully up-to-date)

[...]

PR #106, opened 2023-02-23
Migrate TES configuration to IOptions

BMurri:
Addresses #100 phase 1


src/TesApi.Web/Options/ApplicationInsightsOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// <summary>
+        /// Configuration section.
+        /// </summary>
+        public const string SectionName = "ApplicationInsights";
+
+        /// <summary>
+        /// Account name.
+        /// </summary>
+        public string AccountName { get; set; }
+        // TODO: add additional properties to run without a managed id?
```
MattMcL4475, line 20: `ConnectionString`

https://learn.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core?tabs=netcorenew%2Cnetcore6

Seems like if ConnectionString is specified, it should be used instead.  Thoughts?  Can you add this?

BMurri: Done

PR #110, opened 2023-03-01
Change jobnotfound polly from 10 seconds max to 63 seconds max without flooding the logs

BMurri:


MattMcL4475 (approved):

MattMcL4475:
Should we also add a Delay after adding the job of say, 2 seconds, to reduce retries and log messages?

PR #116, opened 2023-03-07
Mitigate nuget changes for update batch

BMurri:
This is now needed for all github actions


BMurri:
If this action will be available in main, this change should be included there as well

MattMcL4475 (approved):

Issue #72, opened 2023-02-08
Dockerfile-Tes fails to build


colin-nolan:
**Describe the bug**
`src/Dockerfile-Tes` fails to build.

Note: I don't think it's documented anywhere how an image for this service should be built. In lieu of a published image (https://github.com/microsoft/ga4gh-tes/issues/3) I have _assumed_ `./src/Dockerfile-Tes` builds such an image but this may be wrong. Thus perhaps I shouldn't expect this to work if it's for a different use case!

**Steps to Reproduce**
```
$ DOCKER_BUILDKIT=0 docker build -f src/Dockerfile-Tes -t ga4gh-tes src
Sending build context to Docker daemon  1.702MB
Step 1/16 : FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build-env
 ---> 543d806aedf7
Step 2/16 : WORKDIR /app
 ---> Using cache
 ---> fecd7ee10057
Step 3/16 : COPY TesApi.Web/TesApi.Web.csproj TesApi.Web/
 ---> Using cache
 ---> fda3cc58152b
Step 4/16 : COPY Tes/Tes.csproj Tes/
 ---> Using cache
 ---> c7e57afcb7cf
Step 5/16 : RUN dotnet restore TesApi.Web/TesApi.Web.csproj
 ---> Using cache
 ---> 49e6c5ef900e
Step 6/16 : COPY . ./
 ---> 5ceeb23cf77d
Step 7/16 : RUN dotnet publish -c Release -o out --no-restore TesApi.Web/TesApi.Web.csproj
 ---> Running in c80cae778d86
MSBuild version 17.4.1+9a89d02ff for .NET
/usr/share/dotnet/sdk/7.0.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1004: Assets file '/app/CommonUtilities/obj/project.assets.json' not found. Run a NuGet package restore to generate this file. [/app/CommonUtilities/CommonUtilities.csproj]
  Tes -> /app/Tes/bin/Release/net7.0/Tes.dll
The command '/bin/sh -c dotnet publish -c Release -o out --no-restore TesApi.Web/TesApi.Web.csproj' returned a non-zero code: 1
```

**Expected behavior**
The image builds without further knowledge or actions on the developer's side.

[...]

PR #122, opened 2023-03-14
Fixes broken dockerfile

MattMcL4475:
Previously, the dockerfile would throw errors like "The "ResolvePackageAssets" task failed unexpectedly.".   Tested with:

`docker build -f "C:\Users\mattmcl\source\repos\microsoft\CromwellOnAzure\src\ga4gh-tes\src\Dockerfile-Tes" .`

Fixes https://github.com/microsoft/ga4gh-tes/issues/72

### DIFF
--- a/src/Dockerfile-Tes
+++ b/src/Dockerfile-Tes
@@ -5,13 +5,13 @@ FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build-env
 WORKDIR /app
 
 # Restore packages in separate layer
-COPY TesApi.Web/TesApi.Web.csproj TesApi.Web/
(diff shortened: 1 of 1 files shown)


BMurri (approved):
If this actually fixes the issue, and results in a working build, great.

If not, the diagnosis may be incorrect: the build account may not have access to the package source. The fix would be to ensure that the build account is in AD and to give that account access to the package source, and none of these changes are needed.

MattMcL4475:
Thanks Blair, this is to fix it so it works building locally for all external users

BMurri:
For external users, the easier fix is to delete `nuget.config`.

PR #124, opened 2023-03-14
Add logging to determine root causes. This is not expected to remain

BMurri:


MattMcL4475:
If this is not expected to remain, perhaps it's "debug" level info, and we can only write it to the logs when the log level is set to a debug level?  Something like that?

PR #137, opened 2023-03-18
Add IOptions to TES Task cache

MattMcL4475:
Is there a simpler way to do this?

### DIFF
--- a/src/Tes/Repository/ConcurrentDictionaryCache.cs
+++ b/src/Tes/Repository/ConcurrentDictionaryCache.cs
@@ -19,7 +19,7 @@ namespace Tes.Repository
         /// </summary>
         public int MaxCount { get; set; } = DefaultMaxMemoryBytes / DefaultMaxObjectSizeBytes;
 
-        public ConcurrentDictionaryCache(IOptions<TesOptions> options)
+        public ConcurrentDictionaryCache(IOptions<TaskCacheOptions> options)
         {
             MaxCount = options.Value.TesTaskCacheMaxMemoryBytes / options.Value.TesTaskCacheMaxObjectSizeBytes;
         }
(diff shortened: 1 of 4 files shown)


src/Tes/Repository/ConcurrentDictionaryCache.cs
giventocode, line 1: I am sure there is a reason, and I am wondering why not use the existing cache service `IAppCache` instead of creating a new implementation.

src/Tes/Repository/ConcurrentDictionaryCache.cs
```diff
@@ -19,11 +19,16 @@ public class ConcurrentDictionaryCache<T> : ICache<T> where T : class
         /// Total number of items in the cache, defaults to 129,930 tasks (about 2,598 concurrent Mutect2 workflows during scatter).
         /// An ID is a Guid, so also take into account the keysToRemove usage
         /// </summary>
-        public int MaxCount { get; set; } = defaultMaxMemoryBytes / (defaultMaxObjectSizeBytes + Guid.Empty.ToString().Length);
+        public int MaxCount { get; set; } = DefaultMaxMemoryBytes / (DefaultMaxObjectSizeBytes + Guid.Empty.ToString().Length);
```
giventocode, line 22: If greater memory control is what is driving this implementation, it is still worth considering the default `IAppCache/MemoryCache` with an object size limit. The tasks should be approximately of the same size, so a max object limit should correlate closely to desired max memory size. One challenge, however, is you likely require another instance of the cache for this. That can be done easily by creating a factory - returns `IAppCache` instances with the configured values. You can then take a dependency of the factory, that can be injected, instead of cache instance where you need a cache.

giventocode (approved):

PR #139, opened 2023-03-20
Add index to TesTask.Json in PostgreSQL

MattMcL4475:
Stress testing has revealed that getting TesTask's by ID is slow, likely because the ID is within the JSON stored as type `JSONB` in PostgresSQL. This adds a PostgreSQL GIN index to the column.

If testing reveals this is also too slow, will pull out Id and State into their own columns with their own indices


BMurri (approved):

MattMcL4475:
Went with HASH index instead of GIN index

PR #141, opened 2023-03-20
Handle database overloads and return 429

### LINKED ISSUE #228: When the database is overloaded, TES should return 429
**Describe the bug**
When the database is overloaded, backpressure needs to be exerted on the caller

**Steps to Reproduce**
Call Create Task 1 million times in quick succession

**Expected behavior**
Return HTTP status code 429

MattMcL4475:
When the database is overloaded (timeouts, connection pool exhaustion), this PR uses global middleware exception handling to return 429 (too many requests) to the caller.

fixes #228


BMurri (approved):
Looks ok to me, except I'm starting to get eyestrain, so you may wish to wait for @giventocode to weigh in.

src/TesApi.Web/DatabaseOverloadedExceptionMiddleware.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace TesApi.Web
```
BMurri, line 1: I'd suggest moving this file to the `Controllers` folder, since that's the only context in which this code will ever operate.

giventocode (commented):
LGTM!

PR #142, opened 2023-03-21
Parallelize BatchScheduler

### LINKED ISSUE #138: Make TES task scheduling activities reasonably scalable.
**Problem:**
The task update loop (currently the single largest workload TES does) scales in an exponential manner (based on measurements).

**Solution:**
The task update loop should scale on a linear scale. See below for suggested solution.

See microsoft/CromwellOnAzure#497. This issue will replace that issue.

**Additional context**
Currently, the Scheduler service passes each active TesTask to BatchScheduler, which performs various API calls to determine the state of that task in Batch, which it uses to then perform operations on behalf of and/or alter the state of the task. This is performed serially.

The proposed solution is instead to perform the following operations in the following order (assignment of responsibilities TBD):
1. In Autopool mode, query all jobs and filter by TesTaskIds, or in auto-scale mode, query all pools & jobs filtering by known pool ids, capturing all task and compute node states at that time. This will drastically reduce the number of Batch API calls when there are large numbers of active TesTasks, while not overloading the system when there are very few, by taking advantage of using Lists to query Batch state. This is the part that causes the most load in TES when scaling up workloads.
2. For each TesTask, process the retrieved information from step 1 and run each state/task pairing that through the current state machine, except that any operations that can be combined across tasks will be combined (to reduce Azure API loads).
3. Update all tasks in the DB.

Examples of 2. include creating pools/jobs (in auto-scale mode, since they are sharable), retrieving image container credentials, etc. The means of combining operations is TBD.

BMurri:
Fixes #138


src/TesApi.Web/Scheduler.cs
```diff
@@ -125,14 +126,24 @@ private async ValueTask OrchestrateTesTasksOnBatch(CancellationToken stoppingTok
 
             var startTime = DateTime.UtcNow;
 
-            foreach (var tesTask in tesTasks)
+            await Task.WhenAll(await batchScheduler.ProcessTesTasksAsync(tesTasks, stoppingToken).Select(ProcessOrchestrationResult).ToArrayAsync(stoppingToken));
```
MattMcL4475, line 129: how many tasks is this potentially awaiting?  Would it make sense to await batches of 100 of them at a time or something?  Just want to be sure this doesn't await 5000 tasks that causes massive memory load

src/TesApi.Web/AzureProxy.cs
```diff
@@ -290,9 +289,94 @@ public async Task DeleteBatchJobAsync(PoolInformation poolInformation, Cancellat
...
+                jobs = await CachingWithRetriesAzureProxy.asyncRetryPolicy.ExecuteAsync(() =>
+                        batchClient.JobOperations.ListJobs(jobFilter).ToAsyncEnumerable(),
+                        CachingWithRetriesAzureProxy.retryPolicy)
+                    .Where(j => idList.Contains(j.Id))
+                    .ToListAsync(cancellationToken);
+
+                pools = CachingWithRetriesAzureProxy.asyncRetryPolicy.ExecuteAsync(() =>
+                        batchClient.PoolOperations.ListPools(poolFilter).ToAsyncEnumerable(),
+                        CachingWithRetriesAzureProxy.retryPolicy)
+                    .Where(p => idList.Contains(p.Id));
```
MattMcL4475, line 348: Any potential scale concerns here?

src/TesApi.Web/AzureProxy.cs
```diff
@@ -308,18 +392,14 @@ public async Task<AzureBatchJobAndTaskState> GetBatchJobAndTaskStateAsync(TesTas
                 var attemptNumber = 0;
                 CloudTask batchTask = null;
 
-                var jobOrTaskFilter = new ODATADetailLevel
-                {
-                    FilterClause = $"startswith(id,'{tesTask.Id}{BatchJobAttemptSeparator}')",
-                    SelectClause = "*"
-                };
+                var jobOrTaskFilter = new Func<string, string, bool>((id, key) => id.StartsWith(string.Format($"{{0}}{BatchJobAttemptSeparator}", key)));
```
MattMcL4475, line 395: IgnoreCase?

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -361,14 +363,89 @@ await foreach (var cloudPool in GetCloudPools())
...
+                => (listOfTasks, await azureProxy.GetBatchAccountStateAsync(
+                enableBatchAutopool,
+                enableBatchAutopool
+                    ? listOfTasks.Select(t => t.Id)
+                    : GetPools().Select(p => p.Pool.PoolId),
+                cancellationToken));
+        }
+
+        // Adapted from https://stackoverflow.com/a/70026589
+        private static async IAsyncEnumerable<T> WhenEach<T>(IEnumerable<T> items, Func<T, Task> getTask, [EnumeratorCancellation] CancellationToken cancellationToken)
```
MattMcL4475, line 387: do we really need this?

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -361,14 +363,89 @@ await foreach (var cloudPool in GetCloudPools())
...
+            async Task<(List<TesTask> TesTasks, (IReadOnlyDictionary<CloudPool, IReadOnlyList<ComputeNode>> PoolsAndNodes, IReadOnlyDictionary<CloudJob, IReadOnlyList<CloudTask>> JobsAndTasks) State)> GetState(List<TesTask> listOfTasks)
+                => (listOfTasks, await azureProxy.GetBatchAccountStateAsync(
+                enableBatchAutopool,
+                enableBatchAutopool
+                    ? listOfTasks.Select(t => t.Id)
+                    : GetPools().Select(p => p.Pool.PoolId),
+                cancellationToken));
+        }
+
+        // Adapted from https://stackoverflow.com/a/70026589
```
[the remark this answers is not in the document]
BMurri: Mitigated

src/TesApi.Web/BatchScheduler.cs
BMurri: Simpler means that all the tasks are batched and clumped at each stage of processing, which would only increase API retries due to always hammering the same set of resources at the same time and eliminating much (if not most) of the benefit of this effort.

Doing it this way will mean that task updates flow back to the database as soon as they are determined, which will speed up workflows overall and prevent issues related to one task from delaying reporting of all other tasks.

src/TesApi.Web/AzureProxy.cs
BMurri: Mitigated

src/TesApi.Web/AzureProxy.cs
BMurri: Fewer calls will be faster than smaller calls, if all the information is eventually needed.

src/TesApi.Web/Scheduler.cs
BMurri: However many of them there are. The GC is most efficient (and effective) when you load up a bunch of objects and then drop them all, and this basically does that.

Dividing the list into batches by size makes the code more complicated, and actually makes the job of the GC harder, but would mitigate spikes in the memory usage that could result in OutOfMemory errors. Performance metrics is the place to determine if that's a concern, it's not something we can know at this point.

BMurri:
One item left: extend the stopping/cancelation token through the task processing.

BMurri:
With the node runner, we now have events

PR #156, opened 2023-03-23
Refactor and improve PostgreSQL DI

MattMcL4475:
This simplifies the DI and also moves files to TES so that the TriggerService can access it, as well as moves cache warming to a responsibility of the repository


src/Tes/Repository/TesTaskPostgreSqlRepository.cs
```diff
@@ -21,30 +24,21 @@ public class TesTaskPostgreSqlRepository : IRepository<TesTask>
...
-        public TesTaskPostgreSqlRepository(IOptions<PostgreSqlOptions> options, ICache<TesTask> cache)
+        public TesTaskPostgreSqlRepository(IOptions<PostgreSqlOptions> options, ILogger<TesTaskPostgreSqlRepository> logger, ICache<TesTask> cache = null)
         {
             this.cache = cache;
+            this.logger = logger;
             var connectionString = new ConnectionStringUtility().GetPostgresConnectionString(options);
             createDbContext = () => { return new TesDbContext(connectionString); };
             using var dbContext = createDbContext();
             dbContext.Database.MigrateAsync().Wait();
+            WarmCacheAsync().Wait();
```
giventocode, line 41: Not sure this needs to be a blocking operation at start up.  Requests received before the cache is consistent are okay to go the DB the first time. 

Also, there's a service that starts and runs only once in TES and happens on the background - this could be handy if you reconsider the blocking operation.

giventocode (approved):

src/Tes/Repository/TesTaskPostgreSqlRepository.cs
MattMcL4475: Fortunately warming the cache takes less than 2 seconds.  There is a possible race condition if we don't block (need to analyze how likely) if the BatchScheduler updates a TES task to `COMPLETED` _before_ the cache warming is complete, because it will result in the TesTask still in memory with state `RUNNING` and Cromwell never thinking the task is complete until it expires from the cache (30 days) or TES is restarted.

MattMcL4475: I'll implement the above in another PR

PR #160, opened 2023-03-24
Get Application Insights connection string instead of instrumentation key

MattMcL4475:
It's now recommended to use connection strings: https://learn.microsoft.com/en-us/azure/azure-monitor/app/migrate-from-instrumentation-keys-to-connection-strings#new-capabilities


MattMcL4475:
Currently doesn't make a difference

BMurri (approved):

PR #172, opened 2023-03-31
Update schema for Service Info to match official GA4GH Service Info.

### LINKED ISSUE #164: [# 7 TES 1.1 Spec compliance] Update service info to match GA4GH spec
**Describe the bug**
Update service info to match GA4GH spec [serviceinfo spec](https://github.com/ga4gh-discovery/ga4gh-service-info/blob/develop/service-info.yaml#L38)

Required:
        - id
        - name
        - type
        - organization
        - version

Currently only have

`
{
"name": "GA4GH Task Execution Service",
"doc": "",
"storage": [],
"tesResources_backend_parameters": [
"vm_size",
"workflow_execution_identity"
]
}
`

### LINKED ISSUE #249: [#7 TES 1.1 Spec compliance] Validation errors for TesServiceInfo
**Describe the bug**
ServiceInfo does not include the new required properties in the TES 1.1 spec

**Steps to Reproduce**
Run the TES compliance test for TES 1.1

**Expected behavior**
The `response_schema_validation` test case passes.

**Additional context**
Currently, the test case message says: `Response Schema validation failed for GET /service-info`
and the log_messages contains: ` validation errors for TesServiceInfo\nid\n  field required (type=value_error.missing)\ntype\n  field required (type=value_error.missing)\norganization\n  ...`

vsmalladi:
Closes #164
Fixes #249
CoA companion PR microsoft/CromwellOnAzure#686


src/TesApi.Web/Controllers/TaskServiceApi.cs
```diff
@@ -210,13 +210,21 @@ public virtual IActionResult GetServiceInfo()
         {
             var serviceInfo = new TesServiceInfo
```
BMurri, line 211: Since the values in this object never change while the process is running, I'd suggest moving its creation to Startup.cs as a singleton and capture it in this object's constructor.

src/deploy-tes-on-azure/Deployer.cs
```diff
@@ -1105,11 +1104,13 @@ private async Task EnableWorkloadIdentity(ManagedCluster aksCluster, IIdentity m
         {
             settings ??= new();
             var defaults = GetDefaultValues(new[] { "env-00-tes-version.txt", "env-01-account-names.txt", "env-02-internal-images.txt", "env-04-settings.txt" });
+            var currentTime = DateTime.UtcNow;
 
             // We always overwrite the CoA version
             UpdateSetting(settings, defaults, "TesOnAzureVersion", default(string), ignoreDefaults: false);
             UpdateSetting(settings, defaults, "ResourceGroupName", configuration.ResourceGroupName, ignoreDefaults: false);
             UpdateSetting(settings, defaults, "RegionName", configuration.RegionName, ignoreDefaults: false);
+            UpdateSetting(settings, defaults, "DeploymentUpdated", currentTime.ToString("O"), ignoreDefaults: false);
```
BMurri, line 1113: This line should actually set `ignoreDefaults: true`

src/deploy-tes-on-azure/scripts/helm/values-template.yaml
```diff
@@ -19,6 +19,15 @@ config:
...
     prefix: RUNTIME_PARAMETER
+  deployment:
+    organizationName: RUNTIME_PARAMETER
+    organizationUrl: RUNTIME_PARAMETER
+    contactUri: RUNTIME_PARAMETER
+    environment: RUNTIME_PARAMETER
+    created: RUNTIME_PARAMETER
+    updated: RUNTIME_PARAMETER
+  nodeImages:
+    docker: RUNTIME_PARAMETER
```
BMurri, line 30: `nodeImages.docker` is no longer used. It should not be restored. If we're still setting it anywhere else, it should be removed.

src/TesApi.Web/Controllers/TaskServiceApi.cs
```diff
@@ -238,14 +241,14 @@ public virtual async Task<IActionResult> CreateTaskAsync([FromBody] TesTask tesT
         public virtual IActionResult GetServiceInfo()
         {
             var serviceInfo = new TesServiceInfo
```
BMurri, line 243: Remove this. You already have the one you want to use at line 64

BMurri (changes requested):
Also, the updates to `src\deploy-tes-on-azure\scripts\helm\templates\tes-deployment.yaml` to wire what the deployer writes into aksValues.yaml (which originates from `values-template.yaml`) into `src/TesApi.Web/Options/DeploymentOptions.cs` are missing.

I addressed these.

src/deploy-tes-on-azure/Deployer.cs
BMurri: Yes, I know the lines above it have that set to false. There's a rhyme to the madness. This value is different than those others, only sharing the fact that it needs to be rewritten every time.

BMurri: I'm wrong (it's been a very long time since I wrote that code). `false` is correct.

src/TesApi.Web/Controllers/TaskServiceApi.cs
BMurri: I addressed this

src/deploy-tes-on-azure/scripts/helm/values-template.yaml
BMurri: I addressed this

BMurri (approved):

PR #186, opened 2023-04-05
Change auto-scale evaluation interval to 30s

BMurri:


BMurri:
PM clarified that this feature has not yet been implemented.

PR #193, opened 2023-04-12
Fix integration test on linux

jsaun:
Its still unclear to me why this issue only presents when running on linux, maybe the helm process returns faster. Waiting for cert-manager to start + 10 secs fixes it though, waiting for cert-manager alone still results in failures. The bottom issue is some form of deadlock on the tls-secret object, where cert-manager is unable to update because the object was modified while cert-manager was running. 

Other cleanup:
Create namespace was unnecessary. 
process.StandardError.ReadLineAsync() was occasionally throwing exceptions, process.OutputDataReceived += new DataReceivedEventHandler(OutputHandler) seems to be better.


MattMcL4475 (approved):

MattMcL4475:
Thanks for digging deep on this one!

PR #194, opened 2023-04-12
Add BlockBlobDatabase as deployment option

MattMcL4475:


src/Tes/Repository/BlockBlobDatabase.cs
```diff
@@ -0,0 +1,200 @@
...
+                // Delete active if it exists
+                var blobClient1 = container.GetBlobClient(GetActiveBlobNameById(id));
+                var task1 = blobClient1.DeleteIfExistsAsync();
+
+                // Update/create active
+                var blobClient2 = container.GetBlobClient(GetInactiveBlobNameById(id));
+                var task2 = blobClient2.UploadAsync(BinaryData.FromString(json), overwrite: true);
+
+                // Retry to reduce likelihood of one blob succeeding and the other failing
+                await Policy
```
giventocode, line 62: Hmm... I think this retry policy could lead to data loss. Two requests roughly at the same time hit this method, the first fails and the second succeeds (we want the second one to win), the first request will go in retry loop that could overwrite the latest data. 

As an alternative you can consider an optimistic concurrency approach for the update, where you check if an item has changed since you last read it - storage supports this via headers, and etags.  And for the create scenario, you turn overwrite off,  to avoid any race condition.

PR #195, opened 2023-04-12
Trim '?' off SAS Tokens

ashanhol:
Fixes [#638](https://github.com/microsoft/CromwellOnAzure/issues/638)
This PR expands SAS Token properties' setters to trim off any possible leading '?' characters. 

There appears to be one other place a SAS Token is referenced in `GetMappedSasContainerUrlFromWsmAsync()` but that's already trimming a leading `/` caracter


src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1858,10 +1858,16 @@ public async ValueTask<bool> ActionAsync(TesTask tesTask, CombinedBatchTaskInfo
...
+            private string sasToken;
+
             public string AccountName { get; set; }
             public string ContainerName { get; set; }
             public string BlobEndpoint { get; set; }
-            public string SasToken { get; set; }
+            public string SasToken
+            {
+                get => sasToken;
+                set => sasToken = value.TrimStart('?');
```
MattMcL4475, line 1869: add the null-conditional operator in case someone tries to set as null

`set => sasToken = value?.TrimStart('?');`

MattMcL4475 (approved):

PR #196, opened 2023-04-13
Prevent transient CloudException from failing task during pool/job creation

BMurri:


src/TesApi.Web/BatchPool.cs
```diff
@@ -481,19 +481,21 @@ public async ValueTask CreatePoolAndJobAsync(Microsoft.Azure.Management.Batch.Mo
...
+                    }, cancellationToken));
 
                 Configure(pool);
                 _ = _batchPools.AddPool(this);
             }
             catch (AggregateException ex)
             {
                 var exception = ex.Flatten();
-                if (exception.InnerExceptions.Count < 2)
+                if (exception.InnerExceptions?.Count != 1)
```
MattMcL4475, line 498: can you add a comment?  It seems oddly specific.  Only do something if 0 or >1 InnerExceptions ?

Issue #86, opened 2023-02-14
Reduce access level for blobs to read-only for download scenarios


giventocode:
SAS tokens generated for download have access greater than read.

AC:
 - All download operations that generate SAS token for read scenarios must use read only.
 - Must refactor code to facilitate the implementation.

PR #208, opened 2023-04-24
Implement principle of least privilege in SAS token generation

MattMcL4475:
This PR implements the principle of least privilege by only giving SAS tokens the least numbers of permissions they need to perform the given task.  Fixes: https://github.com/microsoft/ga4gh-tes/issues/86


src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
```diff
@@ -85,7 +85,7 @@ public override async Task<bool> IsPublicHttpUrlAsync(string uriString)
         }
 
         /// <inheritdoc />
-        public override async Task<string> MapLocalPathToSasUrlAsync(string path, bool getContainerSas = false)
+        public override async Task<string> MapLocalPathToSasUrlAsync(string path, bool getContainerSas = false, SharedAccessBlobPermissions permissions = SharedAccessBlobPermissions.Read | SharedAccessBlobPermissions.Create | SharedAccessBlobPermissions.Write | SharedAccessBlobPermissions.List)
```
BMurri, line 88: Given the fact that `permissions` are, for all intents and purposes, always provided, while `getContainerSas` is not, I would recommend swapping the order of those two (and please consider making `permissions` required rather then optional).

BMurri (approved):

MattMcL4475:
Already implemented

PR #210, opened 2023-04-27
Fix allowedVmSizes

jsaun:
AllowedVmSizes isn't currently working after some recent refactor. The allowed vm list was being set up in the constructor before the DoOnceAtStartUpService service has had a chance to run and parse the file.


BMurri (commented):
The technique used [here](https://github.com/microsoft/ga4gh-tes/blob/3a2084e6b885851d5b88695814acfb9d7b13b489/src/TesApi.Web/BatchPoolService.cs#L66) affords a way to delay the start of a subsequent background service in order to ensure that all needed initialization has been performed.

When using the `IOptions` pattern, its an anti-pattern to hold onto `IConfiguration` unless your code is actively changing configuration values.

My honest opinion is that `DoOnceAtStartUpService` is an anti-pattern by itself, and the operations performed here should be called from either Program.cs or Startup.cs, but I'm not going to litigate that here.

jsaun:
I'm not quite following how you could use .Wait() in the Startup ConfigureServices method to hold off the batchscheduler singleton until after DoOnceAtStartUpService has run. 

I don't have enough experience with IOptions, why is it an anti-pattern to hold onto IConfiguration. It looks like one of the first examples in Microsoft docs is doing just that: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-7.0#bind-hierarchical-configuration

BMurri:
That first example is explicitly NOT the dependency injection example. The second example is explicitly is.

The idea with the `Wait` is to prevent subsequent services from starting until the required initialization code has completed (so, effectively turning a `Task` return into the equivalent of a `void` return, where the return doesn't happen until the method truly has completed). My suggestion was to apply it in the `DoOnceAtStartUpService` itself if a means of performing the initialization could not be easily performed in the Program or Startup classes.

src/TesApi.Web/AllowedVmSizesService.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace TesApi.Web
+{
+    /// <summary>
+    /// Hosted service that executes one-time set up tasks at start up.
```
MattMcL4475, line 14: should fix this comment to be relevant too

src/TesApi.Web/AllowedVmSizesService.cs
```diff
@@ -0,0 +1,91 @@
...
+    public class AllowedVmSizesService : BackgroundService, IAllowedVmSizesService
+    {
+        private readonly TimeSpan refreshInterval = TimeSpan.FromHours(24);
+        private readonly ILogger logger;
+        private readonly ConfigurationUtils configUtils;
+        private List<string> allowedVmSizes;
+        private Task firstTask;
+
+        /// <summary>
+        /// Hosted service that executes one-time set-up tasks at start up.
```
MattMcL4475, line 25: should fix this comment to be relevant too

MattMcL4475 (approved):

Issue #148, opened 2023-03-22
[Feature] Implement TES Task Runner 


giventocode:
**Overview**
TES dynamically creates a batch script when a new task is submitted. This approach presents several challenges such as:
- Current implementation is hard to maintain and enhance - code that generates code is complex.
  -  New enhancements such as #126 , is challenging to implement considering al possible code paths - ARM, Terra, DRS, etc. 
  -  Robust testing to validate this functionality in isolation is not available and impractical to implement - issue: #62 
- No straightforward way to stream log data as the task is executed #212.
- No feasible way to stream log data as the task is executed.
- Download and upload operations are performed sequentially.
- Download from S3 or GCP sources are hard to implement. 

**Solution**
The proposed solution is to create a new executable that runs on the Batch node that controls download/upload operations as well as the execution of commands.  The following is the AC for the feature:

 - Must enable download and upload operations for all scenarios and resolution approaches.
 - Must remove the dependency with the `blobxfer` and the `docker` containers. 
 - Must determine best way to deploy the executable on the Batch Node.
 - Must allow for the execution of commands on containers. 
 - Consider starting once when the node starts running, and only exiting when the node is deallocated.  Abandon use of Azure Batch task ontology for task execution, and instead have nodes dequeue tasks from a queue (evaluate both Service Bus Queues and Storage Queues).  This will minimize wasted time when the node is running but no task work is being done

Implementation Issues:
- [x] #143 
- [x] #215 
- [x] #216 
- [x] #220 
- [x] #144  
- [x] #145 
- [x] #149  
- [ ] #150 
- [x] #151 
- [x] #245
- [x] #246  
- [x] #279 
- [x] #437 
- [x] #439 
- [x] #440 
- [x] #438 

```[tasklist]
### Tasks
```

MattMcL4475:
Also the "Metadata Requirements" (private link)

PR #211, opened 2023-05-01
TES node runner project

giventocode:
This is the initial PR of the TES node runner.  A few things to call out:

 - This PR includes three new projects, `Tes.Runner` class library with the core implementation, `Tes.Runner.Test` test project for the library, and `Tes.RunnerCLI` a lightweight CLI wrapper for the library and the entry-point for execution.  Checkout [README](https://github.com/microsoft/ga4gh-tes/blob/ja-tes-runner/Tes.RunnerCLI/README.md) for more details.
 - The initial focus is on the download and upload operations so we can migrate those first - checkout  #148 for more details.
 - Contains initial implementation for the Docker execution using the Docker API,  but follow up work is needed to iron out edge-cases and better error handling. 
 - The test project, includes integration tests, however these are disabled (`[Ignore]`) for now as they require `Azurite` which I think is not included in our build pipelines.


Tes.Runner/Executor.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            logger.LogInformation($"Writers:{blobPipelineOptions.NumberOfWriters}");
+            logger.LogInformation($"Readers:{blobPipelineOptions.NumberOfReaders}");
+            logger.LogInformation($"Capacity:{blobPipelineOptions.ReadWriteBuffersCapacity}");
+            logger.LogInformation($"BlockSize:{blobPipelineOptions.BlockSize}");
+        }
+
+        private static async Task<TimeExecutionResult<T>> TimedExecutionAsync<T>(Func<Task<T>> execution)
+        {
+            var sw = new Stopwatch();
```
MattMcL4475, line 143: Stopwatch.StartNew();

MattMcL4475 (approved):

PR #214, opened 2023-05-04
Extend CancellationToken

BMurri:
This extends CancellationToken to all the methods that call asynchronous cancellable methods, along with other minor edits.

Addresses #138

### DIFF
--- a/src/Tes/Repository/IRepository.cs
+++ b/src/Tes/Repository/IRepository.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq.Expressions;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Tes.Repository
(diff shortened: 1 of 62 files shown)


MattMcL4475:
Are there significant risks that this inadvertently causes new bugs ( With such a large PR, that's always a concern)?  What was the reason for doing this, was this necessary for other changes?  Is there more work to be done to completely implement cancellation tokens throughout the codebase?

BMurri:
> Are there significant risks that this inadvertently causes new bugs (with such a large PR, that's always a concern)?

I would be surprised by any bugs introduced with this PR, although I'm certainly not perfect. I intend to run this through its paces as it is here, in this PR, in addition to all the testing I've already done with this in the context of the parallelization solution.

> What was the reason for doing this, was this necessary for other changes?

When I was working on the effort to delete pools, I noticed that during shutdown of the container, operations were often cut off because the cancelations never reached them, and that was without any of the parallelization added, due to the fact that k8 will timeout and force down a container that's "taking too long". Cancelling an operation in progress (especially when you aren't going to ever look at the result) is better than leaving it hanging because the remote API will often cancel when the socket is closed, if it's cancelable. This will help ensure that updates to the images can go as quickly as possible, and will help ensure that remote states (those in other azure resources we're communicating with) will remain in more consistent, expected states.

> Is there more work to be done to completely implement cancellation tokens throughout the codebase?

This was implemented by walking all of the code flows starting from both the API controller and the scheduler background service. I don't believe there's more but I haven't done a final review to ensure that. I will verify this early next week.

PR #224, opened 2023-05-11
Remove remaining references to cromwell-executions

BMurri:


src/TesApi.Web/BatchScheduler.cs
```diff
@@ -447,7 +447,8 @@ private static string StandardizeStartTaskPath(string startTaskPath, string defa
         /// <param name="inputFile"><see cref="Tes.Models.TesInput"/> file</param>
         /// <returns>True if the file is a Cromwell command script</returns>
         private static bool IsCromwellCommandScript(TesInput inputFile)
-            => inputFile.Name.Equals("commandScript");
+            // See https://github.com/broadinstitute/cromwell/blob/17efd599d541a096dc5704991daeaefdd794fefd/supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala#L58
```
ashanhol, line 450: nit: is having a hard link to the line number sustainable to have in the codebase? Could we link to the file and say look at `private val commandScript`?

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -447,7 +447,8 @@ private static string StandardizeStartTaskPath(string startTaskPath, string defa
...
+            => inputFile.Name.Equals("commandScript") && inputFile.Description.EndsWith(".commandScript") && inputFile.Type == TesFileType.FILEEnum;
```
ashanhol, line 451: Why is the name not enough? What else could commandScript be?

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1247,7 +1259,7 @@ private async Task<TesInput> GetTesInputFileUrl(TesInput inputFile, string taskI
             {
                 // Convert file:///account/container/blob paths to /account/container/blob
                 var url = Uri.TryCreate(inputFile.Url, UriKind.Absolute, out var tempUrl) && tempUrl.IsFile ? tempUrl.AbsolutePath : inputFile.Url;
-                inputFileUrl = (await storageAccessProvider.MapLocalPathToSasUrlAsync(url)) ?? throw new TesException("InvalidInputFilePath", $"Unsupported input URL '{inputFile.Url}' for task Id {taskId}. Must start with 'http', '{CromwellPathPrefix}' or use '/accountName/containerName/blobName' pattern where TES service has Contributor access to the storage account.");
+                inputFileUrl = (await storageAccessProvider.MapLocalPathToSasUrlAsync(url)) ?? throw new TesException("InvalidInputFilePath", $"Unsupported input URL '{inputFile.Url}' for task Id {task.Id}. Must start with 'http', '{CromwellPathPrefix}' or use '/accountName/containerName/blobName' pattern where TES service has Contributor access to the storage account.");
```
ashanhol, line 1262: Since this is the same error we're now throwing in filesToUpload, can we make a var for it?

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1197,15 +1208,14 @@ static IEnumerable<string> MungeCleanupScriptForContainerConfig(IEnumerable<stri
...
         /// <returns>List of modified <see cref="TesInput"/> files</returns>
-        private async Task<TesInput> GetTesInputFileUrl(TesInput inputFile, string taskId, List<string> queryStringsToRemoveFromLocalFilePaths)
+        private async Task<TesInput> GetTesInputFileUrl(TesInput inputFile, TesTask task, List<string> queryStringsToRemoveFromLocalFilePaths)
         {
-            if (inputFile.Path is not null
-                && !inputFile.Path.StartsWith($"{CromwellPathPrefix}/", StringComparison.OrdinalIgnoreCase))
+            if (string.IsNullOrWhiteSpace(inputFile.Path) || !inputFile.Path.StartsWith("/"))
             {
-                throw new TesException("InvalidInputFilePath", $"Unsupported input path '{inputFile.Path}' for task Id {taskId}. Must start with '{CromwellPathPrefix}'.");
+                throw new TesException("InvalidInputFilePath", $"Unsupported input path '{inputFile.Path}' for task Id {task.Id}. Must start with '/'.");
```
ashanhol, line 1218: I'm sure you saw, there's a test that checks this string -_-

ashanhol (approved):

src/TesApi.Web/BatchScheduler.cs
BMurri: If the caller is Cromwell, that is sufficient to determine which file is the command script. However, this will false positive for non-cromwell callers that also happen to name their command script "commandScript", and we are using this to try to disambiguate not just the command script, but if it came from cromwell.

src/TesApi.Web/BatchScheduler.cs
BMurri: I didn't see one in this method

src/TesApi.Web/BatchScheduler.cs
BMurri: The message may have happened to end up identical, but it's a different error. It's a different failure reason (that we should have been checking for all along but weren't).

src/TesApi.Web/BatchScheduler.cs
BMurri: For me it's easier to move through the file history easier this way, to see if it's still relevant, rather than searching for something that may have been renamed between the reference and the subsequent access.

PR #232, opened 2023-05-26
Implement download and upload operations in TES scheduler via the TES Task runner

### LINKED ISSUE #144: Implement download and upload operations in TES scheduler via the TES Task runner
**Overview**
The TES Task runner provides the ability to download and upload files using the a SAS URL. The goal of this story is to retrofit the existing batch scheduler implementation to use this functionality. 

AC:
 - Must remove the dependency with `blobxfer` for download/upload.
 - Must define how the TES Task runner will be deployed to the Batch Node. 
   -  A possible approach would consist of deploying the runner and its tasks definition file as  Batch resources.
 - Must use the SAS resolution strategy of `None` - just-in-time SAS resolution is follow up work.
 - Should maintain the existing functionality for DRS.
 - Should maintain the existing approach where the SAS tokens are generated upfront by the TES server.
 - Should consider using the `upload` and `download` commands of the runner. 

 
 DoD:
- Code is merged into main, and all integration tests pass.

BMurri:
fixes #144

I ran both the blobuploadertests and blobdownloadertests and they both passed.

The new process argument `skipMissingSources` is temporary: it will be retired when we only call the node runner once per task, instead of 3 times.

The three new properties of `NodeTask` I expect are also temporary: they should be folded into the metadata requirements as the content of the current `metrics.txt` file is part of that larger effort.

To minimize the effects of this PR, and for simplicity's sake, we are downloading the node task runner for each task. In future work, that will be moved to a permanent start-task operation.

To obtain the self-contained/single-file published build of the node task runner, we are accessing it via a project reference in `TesApi.Web`. We should consider moving that out of the project and into the build pipeline.


src/Tes.Runner/Executor.cs
```diff
@@ -56,14 +56,28 @@ private NodeTask DeserializeNodeTask(string nodeTask)
...
+            }
+
+            return bytesTransfered;
+        }
+
+        private delegate ValueTask<long> TransferIOAsyncDelegate(Channel<byte[]> memoryBufferChannel, bool skipMissingSources);
+
+        public Task<long> UploadOutputsAsync(bool skipMissingSources)
+        {
+            return TransferIOCommonAsync(skipMissingSources, tesNodeTask.OutputsMetricsFormat, UploadOutputsAsync);
```
giventocode, line 77: I am not fully convinced this method is a good abstraction.  At this level, it makes sense to me to keep the  upload and download calls implemented explicitly. Although now they look very similar, they may diverge in the future with other parameters and special scenarios. This implementation makes it harder to maintain as now the call is delegate. I would however create a method to handle the formatting of the metrics, which in reality the only additional thing this method is doing.

src/Tes.Runner/Transfer/MetricsFormatter.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Tes.Runner
+{
+    internal class MetricsFormatter
```
giventocode, line 6: Is the purpose of this class to create  a metrics file with the output of the upload/download? What is the difference from getting this information from the standard out - which we will handle in the future?  It seems like this to be a bit out of scope of the current issue.

src/Tes.Runner/Transfer/PartsProducer.cs
```diff
@@ -33,17 +59,34 @@ public PartsProducer(IBlobPipeline blobPipeline, BlobPipelineOptions blobPipelin
     /// <returns></returns>
     public async Task StartPartsProducersAsync(List<BlobOperationInfo> blobOperations, Channel<PipelineBuffer> readBufferChannel)
```
giventocode, line 60: Now this method has the obfuscated functionality of skipping the missing files - which is not right from the maintainability perspective. This is another complexity that arises from the handling skipping sources at the library level.

src/Tes.RunnerCLI/Commands/CommandFactory.cs
```diff
@@ -7,67 +10,81 @@ internal class CommandFactory
...
+            CreateOption<int>(BlobPipelineOptionsConverter.BlockSizeOption, "Blob block size in bytes", "-b", defaultValue: BlobSizeUtils.DefaultBlockSizeBytes),
+            CreateOption<int>(BlobPipelineOptionsConverter.WritersOption, "Number of concurrent writers", "-w", defaultValue: BlobPipelineOptions.DefaultNumberOfWriters),
+            CreateOption<int>(BlobPipelineOptionsConverter.ReadersOption, "Number of concurrent readers", "-r", defaultValue: BlobPipelineOptions.DefaultNumberOfReaders),
+            CreateOption<int>(BlobPipelineOptionsConverter.BufferCapacityOption, "Pipeline buffer capacity", "-c", defaultValue: BlobPipelineOptions.DefaultReadWriteBuffersCapacity),
+            CreateOption<string>(BlobPipelineOptionsConverter.ApiVersionOption, "Azure Storage API version", "-v", defaultValue: BlobPipelineOptions.DefaultApiVersion)
+        }.AsReadOnly();
+
+        private static readonly Option SkipMissingSourcesOption = CreateOption<bool>(CommandFactory.SkipMissingSources, "Skip missing sources for file transfers", "-s", defaultValue: false); // TODO: remove this option when we incorporate the batch node script logic into this executable
```
giventocode, line 30: Is adding this option the only reason that lead to the refactoring of the CLI?  That seems out of scope of the issue. Specially if the option seems to be temporary.

src/Tes.RunnerCLI/Commands/CommandFactory.cs
```diff
@@ -76,53 +93,55 @@ internal static Command CreateCommand(string optName, string optDescription, par
...
-            foreach (var option in allOptions)
+            foreach (var option in options)
             {
                 cmd.AddOption(option);
             }
 
             return cmd;
         }
 
-        private static List<Option> CreateGlobalOptionList()
```
giventocode, line 91: I thinks method clearly states what are the global options in the CLI, trying to abstract that as collection upstream seems less readable to me.

src/Tes.RunnerCLI/Commands/CommandHandlers.cs
```diff
@@ -23,14 +26,13 @@ internal class CommandHandlers
             {
                 var options = BlobPipelineOptionsConverter.ToBlobPipelineOptions(blockSize, writers, readers, bufferCapacity, apiVersion);
 
-                await ExecuteTransferAsSubProcessAsync(CommandFactory.DownloadCommandName, file, options);
+                await ExecuteTransferAsSubProcessAsync(CommandFactory.DownloadCommandName, file, options, false);
```
giventocode, line 29: This code path should not be executed when using the `upload` or `download` command, so I am not sure why this change here. This is only executed when using the root `exec` command which we are not using now.

src/Tes.RunnerCLI/Commands/HandlerEx.cs
```diff
@@ -0,0 +1,140 @@
...
+{
+    // TODO: This will need to be updated or removed when System.CommandLine is updated.
+
+    /// <summary>
+    /// Extends <see cref="Handler"/>
+    /// </summary>
+    /// <remarks>
+    /// This will need to be updated or removed when System.CommandLine is updated.
+    /// </remarks>
+    internal static class HandlerEx
```
giventocode, line 19: I don't think this is needed. We only have three commands, and likely won't have more in the future. The existing approach to set the handler is simpler in my opinion.

giventocode (changes requested):
Thanks for the draft PR! Most of my comments circle around two high level issues: 
 - Not fully understanding why we need to skip missing files and if that is required, make the implementation upstream in the stack wherever the node task definition is created instead of down the weeds of the runner library.
 - Refactoring and functionality that seems out of scope for the issue.

src/Tes.Runner/Transfer/BlobUploader.cs
```diff
@@ -82,9 +82,10 @@ public override async ValueTask<int> ExecuteReadAsync(PipelineBuffer buffer)
         /// </summary>
         /// <param name="lengthSource">File path</param>
         /// <returns>File size in number of bytes</returns>
-        public override Task<long> GetSourceLengthAsync(string lengthSource)
+        public override Task<long?> GetSourceLengthAsync(string lengthSource)
         {
-            return Task.FromResult((new FileInfo(lengthSource)).Length);
+            var file = new FileInfo(lengthSource);
+            return Task.FromResult<long?>(file.Exists ? file.Length : null);
```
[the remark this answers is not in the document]
BMurri: FileInfo explicitly does NOT throw if the file is missing. Instead. the Exists property is False.

src/Tes.RunnerCLI/Commands/CommandFactory.cs
```diff
@@ -7,67 +10,81 @@ internal class CommandFactory
...
+            rootCommand.AddOption(CreateOption<Uri>(CommandFactory.DockerUriOption, "local docker engine endpoint", "-u", defaultValue: DefaultDockerUri));
+
+            foreach (var option in GlobalOptions)
             {
                 rootCommand.AddGlobalOption(option);
             }
 
-            rootCommand.AddOption(dockerUriOption);
+            // GetOptionByName() expects that the "globalOptions" parameter contain only inherited global options. Thus, we don't pass any since we're the root.
+            var fakedGlobalList = Enumerable.Empty<Option>().ToList();
```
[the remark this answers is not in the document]
BMurri: I had eliminated the need for it, then forgot to remove it.

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1122,36 +1105,58 @@ private async Task<CloudTask> ConvertTesTaskToBatchTaskAsync(string taskId, TesT
...
 
+            var uploadMetricsScriptPath = $"/{storageUploadPath}/{UploadMetricsScriptFileName}";
+            var uploadMetricsScriptContent = new NodeTask
+            {
+                Outputs = new List<FileOutput>() { new FileOutput { FullFileName = metricsName, TargetUrl = metricsUrl.ToString(), SasStrategy = SasResolutionStrategy.None } }
+            };
+
             sb.AppendLinuxLine($"write_ts DownloadStart && \\");
-            sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} {MountBlobxferScriptAndMetrics(DownloadFilesScriptFileName)} --entrypoint=/bin/sh {blobxferImageName} /{DownloadFilesScriptFileName} && \\");
+            sb.AppendLinuxLine($"tRunner download --file {DownloadFilesScriptFileName} --blockSize {blobxferChunkSizeBytes:D} --bufferCapacity {blobxferOneShotBytes:D} && \\");
```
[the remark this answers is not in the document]
BMurri: Why is it a command-line option then?

src/Tes.Runner/Transfer/MetricsFormatter.cs
BMurri: We need to maintain the current metrics file while using this binary to replace blobxfer, which makes it very much a part of this issue. This class accomplishes that. In the future you mention, this class can then go away since it's functionality will be replaced by that future solution.

src/Tes.RunnerCLI/Commands/CommandFactory.cs
BMurri: Yes, but as discussed we have a different route to solve this.

src/Tes.RunnerCLI/Commands/HandlerEx.cs
BMurri: This is needed when adding more options and/or arguments to any command such that the total number, including globals, sums past 8, unless you want to lose the pattern you are already using. This simply extends the same pattern already in the third party library to work for more than 8 symbols per command. When I added to the commands, it became necessary. Since it's already written and tested, we can either leave it in if we end up needing it, we can remove the file from Compile so that it's not included until we realize we do need it, or we can dump it and recreate it later when/if that becomes necessary.

src/Tes.RunnerCLI/Commands/CommandFactory.cs
BMurri: The collection is named `GlobalOptions`. Seems very readable to me.

src/Tes.Runner/Storage/ResolutionPolicyHandler.cs
giventocode, line 1: This makes sense, thanks for the simpler implementation of skipping missing files.

src/Tes.Runner/Storage/ResolutionPolicyHandler.cs
```diff
@@ -56,29 +69,29 @@ public class ResolutionPolicyHandler
...
 
-    private async Task<UploadInfo> CreateUploadInfoWithStrategyAsync(FileOutput output)
+    private static async Task<UploadInfo> CreateUploadInfoWithStrategyAsync(FileOutput output)
     {
-        var uri = await ApplySasResolutionToUrlAsync(output.TargetUrl, output.SasStrategy);
+        var uri = await ApplySasResolutionToUrlAsync(output.TargetUrl, output.SasStrategy ?? SasResolutionStrategy.None);
 
-        if (string.IsNullOrEmpty(output.FullFileName))
-        {
-            throw new ArgumentException("A task output is missing the full filename. Please check the task definition.");
```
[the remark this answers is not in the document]
BMurri: The check is moved, not removed. See line 40.

src/Tes.Runner/Storage/ResolutionPolicyHandler.cs
```diff
@@ -56,29 +69,29 @@ public class ResolutionPolicyHandler
...
-        }
+        return new UploadInfo(ExpandEnvironmentVariables(output.FullFileName!), uri);
```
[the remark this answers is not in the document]
BMurri: Since most of the uploads/downloads happen outside of the executor's view of the filesystem, the files that go into/come from the executor need to be prefixed. But, when creating the file, there's no way to know exactly what the prefix is, but there is an environment variable that gets us 99% of the way there.

src/Tes.RunnerCLI/Commands/CommandFactory.cs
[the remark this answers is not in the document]
BMurri: Generating three otherwise identical copies of the same object always opens up the possibility that something on some update to the third party library being called will go sideways, because that is unexpected and outside of the given spec. It also generates non-sensical command-line help (with lines repeated all over the place). The way I left it resolves both of those issues.

src/CommonUtilities/Models/NodeTask.cs
```diff
@@ -12,20 +12,24 @@ public class NodeTask
...
+        public string? InputsMetricsFormat { get; set; }
+        public string? OutputsMetricsFormat { get; set; }
     }
 
     public class FileOutput
     {
         public string? FullFileName { get; set; }
         public string? TargetUrl { get; set; }
-        public SasResolutionStrategy SasStrategy { get; set; }
+        public SasResolutionStrategy? SasStrategy { get; set; }
```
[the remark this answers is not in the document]
BMurri: I agree it shouldn't be, but the json serializer does not agree (it's perfectly willing to set non-nullable properties to null in my experience). I'll verify/throw elsewhere if it's not set.

src/Tes.Runner/Storage/ResolutionPolicyHandler.cs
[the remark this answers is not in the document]
BMurri: Unless now is the time to start inserting details of the runtime compute node environment we use to run the tasks (I was directed not to do so at this stage, leaving it in the hands of TES as it is today).

src/Tes.Runner/Storage/ResolutionPolicyHandler.cs
```diff
@@ -56,29 +69,29 @@ public class ResolutionPolicyHandler
...
     {
-        var uri = await ApplySasResolutionToUrlAsync(input.SourceUrl, input.SasStrategy);
-
         if (string.IsNullOrEmpty(input.FullFileName))
         {
             throw new ArgumentException("A task input is missing the full filename. Please check the task definition.");
         }
 
-        return new DownloadInfo(input.FullFileName, uri);
+        var uri = await ApplySasResolutionToUrlAsync(input.SourceUrl, input.SasStrategy ?? SasResolutionStrategy.None);
```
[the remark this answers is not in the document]
BMurri: Fixed

src/Tes.Runner/Storage/ResolutionPolicyHandler.cs
```diff
@@ -56,29 +69,29 @@ public class ResolutionPolicyHandler
...
+
+        return new DownloadInfo(ExpandEnvironmentVariables(input.FullFileName), uri);
     }
 
-    private async Task<UploadInfo> CreateUploadInfoWithStrategyAsync(FileOutput output)
+    private static async Task<UploadInfo> CreateUploadInfoWithStrategyAsync(FileOutput output)
     {
-        var uri = await ApplySasResolutionToUrlAsync(output.TargetUrl, output.SasStrategy);
+        var uri = await ApplySasResolutionToUrlAsync(output.TargetUrl, output.SasStrategy ?? SasResolutionStrategy.None);
```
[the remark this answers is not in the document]
BMurri: Fixed

src/Tes.Runner/Storage/ResolutionPolicyHandler.cs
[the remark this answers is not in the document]
BMurri: A thought just occurred to me. Would it make more sense to add a "prefix" property to the nodetask object, and let that be the environmentally-resolved prefix, rather than prefix all of the TES paths with a prefix containing an environment variable? If you agree, I'll do that first thing in the morning.

src/Tes.RunnerCLI/Commands/CommandFactory.cs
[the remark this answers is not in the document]
giventocode: Thanks for providing additional insights into the change. This context is not evident from the changes. 

There are three commands in the cli, a root (exec) and two children upload and download. All options for the children commands apply to the root as the execution performs download and upload operations. The root however, has one additional option - docker API URL. The help output for all commands depicts this behavior, so I still don't know what other functionality we are adding or fixing. 

Stepping back a bit, note that I am not suggesting the original implementation is correct or does not require improvement. My overall point is these changes are tangential at best to the issue (even more after the skipping option is not needed), and it is challenging to perform a helpful review without a context of why they are required. In a practical sense, as a reviewer, I have to fall back to consider these as just stylistic changes and not internalize the implementation. This dramatically increases the chances that I may miss helpful feedback. To help the reviewer be more effective, tangential changes must be separate issues, where the context of why they are needed is clearly described and discussed.

src/Tes.Runner/Transfer/MetricsFormatter.cs
giventocode: Sounds good.

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1122,36 +1091,56 @@ private async Task<CloudTask> ConvertTesTaskToBatchTaskAsync(string taskId, TesT
...
+            sb.AppendLinuxLine($"./{NodeTaskRunnerFilename} download --file {DownloadFilesScriptFileName} && \\");
             sb.AppendLinuxLine($"write_ts DownloadEnd && \\");
             sb.AppendLinuxLine($"chmod -R o+rwx $AZ_BATCH_TASK_WORKING_DIR/wd && \\");
             sb.AppendLinuxLine($"export TES_TASK_WD=$AZ_BATCH_TASK_WORKING_DIR/wd && \\");
             sb.AppendLinuxLine($"write_ts ExecutorStart && \\");
             sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} --entrypoint= --workdir / {executor.Image} {executor.Command[0]}  \"{string.Join(" && ", executor.Command.Skip(1))}\" && \\");
             sb.AppendLinuxLine($"write_ts ExecutorEnd && \\");
             sb.AppendLinuxLine($"write_ts UploadStart && \\");
-            sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} {MountBlobxferScriptAndMetrics(UploadFilesScriptFileName)} --entrypoint=/bin/sh {blobxferImageName} /{UploadFilesScriptFileName} && \\");
+            sb.AppendLinuxLine($"./{NodeTaskRunnerFilename} upload --file {UploadFilesScriptFileName} && \\");
```
giventocode, line 1109: A possible simpler implementation would be to create a single node task definition file with the inputs and outputs defined. The download/upload commands will only consider the relevant parts of the definition. Also, if you name the file `TesTask.json` you don't even need to set the `--file` option.

src/Tes.RunnerCLI/Commands/CommandFactory.cs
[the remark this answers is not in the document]
BMurri: Moving this into a separate PR

src/TesApi.Web/BatchScheduler.cs
BMurri: Since we are calling this three times instead of two, and because we are calling this as a replacement for blobxfer/wget, not for its ultimate purpose, it makes sense to have the files replace the upload/download scripts for now, to continue to preserve as much of the current structure as possible. When the node task runner agent takes over the primary responsibility for everything this method currently does, that will be the moment to create one file and name it `TesTask.json`

ashanhol:
[pre-PR reading note] 
Based on your description it sounds like there's a couple of follow up items that will come out of this PR around removing or stabilizing functions. Have you created issues for those?

src/Tes.RunnerCLI/Commands/CommandHandlers.cs
```diff
@@ -47,7 +48,9 @@ private static async Task ExecuteNodeContainerTaskAsync(FileInfo file, Uri docke
         {
             try
             {
-                var executor = new Executor(file.FullName, options);
+                var nodeTask = DeserializeNodeTask(file.FullName);
```
giventocode, line 51: Do we need to wrap this with 'Environment.ExpandEnvironmentVariables'? I am not sure if the CLI will pass an expanded path here?

giventocode (approved):

src/Tes.RunnerCLI/Commands/CommandHandlers.cs
BMurri: When the command line option creates the FileInfo object, it is parsed based on the current directory if it was not already and absolute path. That is automatically expanded by the FileInfo object when accessing the FullName property. Since we call the executor in the same directory as the file was dropped by Azure Batch, it expands correctly, There's no environment variable added to this argument in BatchScheduler since everything at this stage is all in that one directory.

Issue #235, opened 2023-06-01
Root hash implementation for upload operations via the runner


giventocode:
**Overview**

Call caching requires a checksum for blobs in Azure storage. Azure storage does not calculate a file MD5 checksum if a blob is uploaded in multiple blocks. Calculating file MD5 on the client significantly impacts upload performance as it would either serialize multi-block uploads or block the operation until the sequential computation of the hash completes in the background. 

As an alternative, we proposed a checksum implementation based on a hash of block hashes - a root hash of a hash list. The root hash would be obtained by dividing a file into consecutive blocks of a fixed size, computing the MD5 hash (32-character hexadecimal string) for each block, sorting them in by the block's ordinal position in ascending order, concatenating them together, and finally computing the MD5 hash (32-character hexadecimal string) of the concatenated result. 

AC:
- Must set the root hash value as metadata property for the blob when committing the block list.
- Should consider the performance impact of the implementation and minimize any synchronization.

PR #236, opened 2023-06-01
[235] Initial implementation of the root hash for uploads

giventocode:
Closes: #235

This PR includes the initial implementation of root hash (hash of block hashes) for the uploads.
- A hash-list of MD5 checksums is created for each block in the transfer.
- Upon completion, the MD5 checksum is calculated from the hash-list. 
- The resulting MD5 checksum value is set a metadata property for the blob. 

Please note that there are follow-up items (in other issues):
- Use 4MiB as the block size for the hash.
- Dynamically set the block size depending on the source file for uploads.  
- Create a script to calculate the root hash for the file locally.


src/Tes.Runner/Transfer/BlobUploader.cs
```diff
@@ -119,7 +138,7 @@ public async Task<long> UploadAsync(List<UploadInfo> uploadList)
 
         private static void ValidateUploadList(List<UploadInfo> uploadList)
         {
-            ArgumentNullException.ThrowIfNull(uploadList);
+            ArgumentNullException.ThrowIfNull(uploadList, nameof(uploadList));
```
BMurri, line 141: This is not needed. The parameter name in the resulting exception defaults to the string used in the expression, which happens to be exactly the same as what you are passing for the parameter name, It's basically TMI for no benefit.

src/Tes.Runner/Transfer/ProcessedPartsProcessor.cs
```diff
@@ -62,15 +69,25 @@ public ProcessedPartsProcessor(IBlobPipeline blobPipeline)
...
+        {
+            readBufferChannel.Writer.Complete();
+        }
 
         logger.LogInformation("All parts were successfully processed.");
 
         return totalBytes;
     }
 
+    private string GetRootHashIfProviderIsSet(IHashListProvider hashListProvider)
```
BMurri, line 86: This name is misleading. It communicates that if the provider is not set (aka is null) it will not get the root hash. Instead, it would throw an exception because there's no conditional logic here.

src/Tes.Runner/Transfer/Md5HashListProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Tes.Runner.Transfer
+{
+    public class Md5HashListProvider : IHashListProvider
+    {
+        private readonly ILogger logger = PipelineLoggerFactory.Create<Md5HashListProvider>();
+        private readonly ConcurrentDictionary<int, string> hashesDictionary = new();
+
+        public ConcurrentDictionary<int, string> HashList => hashesDictionary;
```
BMurri, line 16: Consider exposing a read-only snapshot instead of the directory object itself here. This isn't something who's content should be changed outside of this class.

BMurri (commented):
Consider using an obviously temporary name for the hash metadata instead of what looks like it might be a final name until the hash block size is fixed unless this will not be shipping soon, as the final hashes will vary based on the hash block size for the same file and once the metadata is being used for either integration or production workflows, it can't be changed without potentially undesirable side effects.

src/Tes.Runner/Transfer/BlobUploader.cs
giventocode: Not sure I agree here. I added the name mostly, because I've seen this pattern in most places in the code base.  If anything, this will result in a consistent implementation.

src/Tes.Runner/Transfer/ProcessedPartsProcessor.cs
giventocode: I agree and will change it.

src/Tes.Runner/Transfer/Md5HashListProvider.cs
giventocode: No need to do this. This property is only use for testing, and must be public because tests are in another project.

giventocode:
> Consider using an obviously temporary name for the hash metadata instead of what looks like it might be a final name until the hash block size is fixed unless this will not be shipping soon, as the final hashes will vary based on the hash block size for the same file and once the metadata is being used for either integration or production workflows, it can't be changed without potentially undesirable side effects.

The metadata name is temp. The release name will include the block size (4MiB) in its name.

src/Tes.Runner/Transfer/Md5HashListProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+            var data = Encoding.UTF8.GetBytes(stringBuilder.ToString());
+
+            var rootHash = CreateBlockMd5CheckSumValue(data, data.Length);
+
+            logger.LogInformation($"Root Hash: {rootHash}");
+
+            return rootHash;
+        }
+
+        private static string CreateBlockMd5CheckSumValue(byte[] buffer, int length)
```
MattMcL4475, line 50: "int length" isn't necessary since can just use buffer.Length

MattMcL4475 (approved):

src/Tes.Runner/Transfer/Md5HashListProvider.cs
giventocode: yeah, this method will be refactored when I implement the 4MiB block for hashes, at that point `length` will be 4MiB, which will be different than `buffer. Length`.

PR #239, opened 2023-06-02
Share global options instead of recreating them multiple times

### LINKED ISSUE #238: Address issues with sharing Options from System.CommandLine across Commands
Generating three otherwise identical copies of the same object always opens up the possibility that something on some update to the third party library being called will go sideways, because that is unexpected and outside of the given spec. It also generates non-sensical command-line help (with lines repeated all over the place).

BMurri:
fixes #238


src/Tes.RunnerCLI/Commands/CommandFactory.cs
```diff
@@ -79,53 +85,55 @@ internal static Command CreateCommand(string optName, string optDescription, par
...
-            throw new InvalidOperationException("Invalid option");
+            throw new InvalidOperationException("Invalid option (name not found)");
+
+            // TODO: Future versions of System.CommandLine are expected to make the flag public, but it appears that it will also be renamed. So we can then get rid of the reflection.
+            static bool IsOptionGlobal(Option option)
+                => (bool)option.GetType().GetProperty("IsGlobal", System.Reflection.BindingFlags.Instance | System.Reflection.BindingFlags.NonPublic)!.GetValue(option)!;
         }
 
-        private static Option<T> CreateOption<T>(string name, string description, string alias, bool required = false, object? defaultValue = null)
+        private static Option<T> CreateOption<T>(string name, string description, string alias, bool required = false, object? defaultValue = null, System.CommandLine.Parsing.ParseArgument<T>? parse = null, Action<Option<T>>? configureOption = null)
```
giventocode, line 129: I don't see any call to CreateOption that uses the last two arguments.

giventocode (approved):
lgtm, one nit comment.

PR #241, opened 2023-06-02
Snap to specific dot net to fix dotnet format

yuliadub:
our issue that tracks this: https://github.com/microsoft/ga4gh-tes/issues/242

explanation regarding why we need the global.json, apparently just specifying it in the step in github action doesnt actually do anything: https://github.com/actions/setup-dotnet#usage


BMurri (approved):
The downside here is that this will force the entire build to run under 7.0.203, which has been superseded by 7.0.302 (in fact, upgrading VS 2022 will replace 203 with 302). I don't know if we will run into any issues with release processes. This should be short-term only (until dotnet fixes their stuff).

BMurri:
I don't know how to rescind my approval, but this isn't building in the pipeline because 7.0.203 has already been blacklisted for releasing new code.

yuliadub:
@BMurri adusted this so the global.json is only used for the github action, not for the project (it lives in the format directory). Let me know if you think this is a reasonable alternative for now.

yuliadub:
> I don't know how to rescind my approval, but this isn't building in the pipeline because 7.0.203 has already been blacklisted for releasing new code.

I had to look this up too, so putting here for the future: https://stackoverflow.com/questions/63910871/how-could-i-remove-my-approval-from-a-pull-request-on-github

BMurri:
> @BMurri adusted this so the global.json is only used for the github action, not for the project (it lives in the format directory). Let me know if you think this is a reasonable alternative for now.

Perfect. Ship it:tm:

PR #251, opened 2023-06-07
Implement backend parameter internal_path_prefix

### LINKED ISSUE #243: Create new configuration setting "TesInternalPath" for TES internal files
**Problem:**
Currently, TES hardcodes the internal TES files to the "tes-internal" container in the default storage account.  This is problematic for Terra, where a workspace only has a single container.

**Solution:**
Add a new configuration setting for TES called "TesInternalPath", that specifies the container and blob prefix for all TES internal files.  Valid examples:

/tes-internal
/tes-internal?sas
tes-internal
/mycontainer/tesinternal
/mycontainer/tesinternal?sas
/mycontainer/tesinternal/
mycontainer/tesinternal/
mycontainer/tesinternal/?sas

The first part should always be interpreted as the container name.  A container SAS may be optionally included.

BMurri:
fixes #243

### DIFF
--- a/src/TesApi.Web/BatchScheduler.cs
+++ b/src/TesApi.Web/BatchScheduler.cs
@@ -415,9 +415,11 @@ namespace TesApi.Web
         private static string GetCromwellExecutionDirectoryPath(TesTask task)
             => GetParentPath(task.Inputs?.FirstOrDefault(IsCromwellCommandScript)?.Path.TrimStart('/'));
 
-        private static string GetStorageUploadPath(TesTask task, string storageAccountName)
+        private string GetStorageUploadPath(TesTask task)
         {
-            return $"{storageAccountName}/{TesExecutionsPathPrefix.TrimStart('/')}/{task.Id}";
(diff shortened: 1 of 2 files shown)


src/TesApi.Web/BatchScheduler.cs
```diff
@@ -415,9 +415,11 @@ public void ClearBatchLogState()
         private static string GetCromwellExecutionDirectoryPath(TesTask task)
             => GetParentPath(task.Inputs?.FirstOrDefault(IsCromwellCommandScript)?.Path.TrimStart('/'));
 
-        private static string GetStorageUploadPath(TesTask task, string storageAccountName)
+        private string GetStorageUploadPath(TesTask task)
```
ashanhol, line 418: Can you add a function descriptor comment? I'm able to follow because I just worked on this code/read the github issue but I'd be pretty confused as to why we're doing this if I was just reading the code.

ashanhol (approved):

src/TesApi.Web/BatchScheduler.cs
BMurri: https://github.com/microsoft/ga4gh-tes/issues/253

Issue #246, opened 2023-06-06
Use 4MiB block size for md5 hash list.


giventocode:
*Overview*
The runner uses a variable block size for upload and download operations.  The initial implementation of the root hash value for file uploads uses the variable block size.  A variable block size complicates the implementation of local or external validation process that must produce the root hash value. To simplify and avoid this challenge we want to use a fixed value of 4MiB.

AC:
 - The block size for hash list must set to 4MiB
 - Must change the metadata property containing the hash root value to `md5_4mib_hashlist_root`
 - Must validate that the block size is a multiple of 4MiB. 
 - Must include a sample bash script to compute the root hash locally using a 4MiB block size.

PR #252, opened 2023-06-07
[246] 4MiB hash list for uploads

giventocode:
Closes: #246 

This PR includes the following:
 - The block/chunk size of each item in the hash list is 4MiB. 
 - Introduced validation to only allow multiples of 4MiB as valid block sizes.
 - The root hash is set in the blob property: `md5_4mib_hashlist_root_hash`
 - Added bash script to generate the root hash locally `/scripts/rootHash.sh`


BMurri (approved):
For some reason I thought we were doing a hash of hashes but I see we are doing a concatenation of hashes. Not necessarily a problem, I was surprised

giventocode:
> For some reason I thought we were doing a hash of hashes but I see we are doing a concatenation of hashes. Not necessarily a problem, I was surprised

It is a hash of hashes. The source of the final hash (root hash) is a list of hashes (concatenated, without a separator) of chunks of 4MiB.

src/Tes.RunnerCLI/scripts/rootHash.sh
```diff
@@ -0,0 +1,21 @@
+﻿#!/bin/bash
+
+filename=$1
+blocksize=4194304 # 4 MiB
```
MattMcL4475, line 4: You could expose it as a parameter with a default value:

`blocksize=${2:-4194304}`

(note: the syntax is weird, it's not a negative number)

src/Tes.RunnerCLI/scripts/rootHash.sh
```diff
@@ -0,0 +1,21 @@
...
+    exit 1
+fi
+
+filesize=$(stat -c%s "$filename")
+numblocks=$(((filesize + blocksize - 1) / blocksize))
+
+rm .hashList
+
+for ((i=0; i<$numblocks; i++)); do
+    (dd if="$filename" bs=$blocksize skip=$i count=1 status=none | md5sum | tr -d '\n-' | tr -d ' ') >> .hashList
```
MattMcL4475, line 17: do you think users really want a local file created (could just do this with a variable).  If they do, then why is it hidden? I would think either don't use a local file, or, if you do, it shouldn't be a hidden file (starting with a period)

MattMcL4475 (approved):

src/Tes.RunnerCLI/scripts/rootHash.sh
```diff
@@ -0,0 +1,21 @@
+﻿#!/bin/bash
```
MattMcL4475, line 1: this file should probably be named "roothash.sh", since in Linux, lower case file names are canonical and it's case sensitive

PR #257, opened 2023-06-13
Update issue template for code dependencies

MattMcL4475:


ashanhol (approved):
You should also modify the issue template in CoA imo

Issue #258, opened 2023-06-13
Consider CPU core count when optimizing the number of threads for the readers and writers


giventocode:
**Overview**
The current optimizer does not consider the CPU when calculating the number of threads for the readers and writers. This leads to high values when there is low CPU count and enough memory, which affect the stability of the transfer operations.

*AC*
 - Must use a function that considers the CPU count as a factor.
 - Must max the thread count with the current maximum value.

PR #259, opened 2023-06-14
Performance testing Fixes and Enhancements

giventocode:
Closes: #258 , #264

This PR includes:
 - new logic in `PipelineOptionsOptimizer` to calculate the number of threads for the read and write operations using the CPU core  count. The new logic sets the number threads at 10 per core with a maximum of 90.
 - Cancellation of channel read/write operations when worker thread fail.
 - Added timeout when reading from a stream to work around this [issue:](https://github.com/dotnet/runtime/issues/36822)


src/Tes.Runner/Transfer/PartsProducer.cs
```diff
@@ -31,6 +31,7 @@ public PartsProducer(IBlobPipeline blobPipeline, BlobPipelineOptions blobPipelin
     /// </summary>
     /// <param name="blobOperations">A list of <see cref="BlobOperationInfo"/>></param>
     /// <param name="readBufferChannel">Channel where are parts are written</param>
+    /// <param name="cancellationToken"></param>
     /// <returns></returns>
     public async Task StartPartsProducersAsync(List<BlobOperationInfo> blobOperations, Channel<PipelineBuffer> readBufferChannel)
```
BMurri, line 36: The method is documented to receive a CancellationToken as the third argument, but the code does not reflect that. If the argument isn't needed, please remove the corresponding `param name`

BMurri (changes requested):
Looks good from this view. If it's doing well in the stress test, I'll approve it then.

src/Tes.Runner/Transfer/BlobBlockApiHttpUtils.cs
```diff
@@ -170,6 +186,15 @@ private static async Task<int> ExecuteHttpRequestAndReadBodyResponseImplAsync(Pi
 
             HandleHttpRequestException(status, ex);
         }
+        catch (TaskCanceledException ex)
```
BMurri, line 189: Even better would be `OperationCanceledException`

src/Tes.Runner/Transfer/BlobBlockApiHttpUtils.cs
```diff
@@ -99,22 +99,38 @@ private static async Task<HttpResponseMessage> ExecuteHttpRequestImplAsync(Func<
...
+                response.EnsureSuccessStatusCode();
+            }
+            catch (HttpRequestException ex)
+            {
+                var status = response?.StatusCode;
 
-            HandleHttpRequestException(status, ex);
+                HandleHttpRequestException(status, ex);
+            }
+            catch (TaskCanceledException ex)
```
BMurri, line 114: Same as below

BMurri (changes requested):

BMurri (approved):

PR #261, opened 2023-06-14
[245] Wildcard support and refactoring of runner task outputs and inputs expansion logic 

### LINKED ISSUE #245: [#148 Implement TES Task Runner] Implement upload directory handling via an OperationsResolver in the library
Overview
The TES Task runner provides the ability to upload files from a directory (#232). The goal of this story is to retrofit that functionality into something that will be maintainable and accommodate changes to TES specs over time.

AC:
 * Must be unit testable and must include unit tests
 * Must determine if current path is file or directory if that is not supplied (TES 1.1 spec)
 * Must filter non-existent optional files
 * Must implement #8
 * Should be called from `ResolutionPolicyHandler`
 * Consider naming the new class `UploadOperationsResolver`

DoD:
 * Code is merged into main, and all integration tests pass.

giventocode:
Closes #245
Addresses #8

This PR includes:
 - A new class: `FileOperationResolver` that performs the resolution process for inputs and outputs of the runner task. 
   -  The resolution process consists of: 
      - Expanding, directories or paths with wildcards (if a `path_prefix` is provided).
         - Wildcard support is limited to the functionality of the standard .NET library (`Directory.GetFiles()`).
      - Apply the SAS resolution strategy for each expanded item.
- Additional changes:
  - Renamed, `FullFileName` property to `Path` in the runner task model (`NodeTask`), to better align with the TES spec and its semantics.
  - Added the property `PathPrefix` to the `TestOut` model in the Web API - 1.1 prerequisite.

### DIFF
--- a/src/CommonUtilities/Models/NodeTask.cs
+++ b/src/CommonUtilities/Models/NodeTask.cs
@@ -19,7 +19,7 @@ namespace Tes.Runner.Models
 
     public class FileOutput
     {
-        public string? FullFileName { get; set; }
+        public string? Path { get; set; }
         public string? TargetUrl { get; set; }
         public SasResolutionStrategy? SasStrategy { get; set; }
         public FileType? FileType { get; set; }
@@ -28,7 +28,7 @@ namespace Tes.Runner.Models
 
     public class FileInput
     {
-        public string? FullFileName { get; set; }
+        public string? Path { get; set; }
         public string? SourceUrl { get; set; }
         public SasResolutionStrategy? SasStrategy { get; set; }
     }
--- a/src/CommonUtilities/Models/NodeTask.cs
+++ b/src/CommonUtilities/Models/NodeTask.cs
@@ -23,7 +23,7 @@ namespace Tes.Runner.Models
         public string? TargetUrl { get; set; }
         public SasResolutionStrategy? SasStrategy { get; set; }
         public FileType? FileType { get; set; }
-        public string? PathPrefix { get; set;}
+        public string? PathPrefix { get; set; }
     }
 
     public class FileInput
--- a/src/CommonUtilities/Models/NodeTask.cs
+++ b/src/CommonUtilities/Models/NodeTask.cs
@@ -23,7 +23,7 @@ namespace Tes.Runner.Models
         public string? TargetUrl { get; set; }
         public SasResolutionStrategy? SasStrategy { get; set; }
         public FileType? FileType { get; set; }
-        public bool? Required { get; set; }
+        public string? PathPrefix { get; set;}
     }
 
     public class FileInput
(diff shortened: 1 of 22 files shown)


BMurri:
If wildcards are not provided, pathprefix should be ignored. If output is a directory and there are no wildcards, the directory's path should be used as the pathprefix. The question is whether to figure this out in the node agent or in TES itself.

Either way, this code doesn't work as intended.

giventocode:
> If wildcards are not provided, pathprefix should be ignored. 

I don't see why this is necessary. The implementation assumes that if a pathprefix is set, then pattern search is intended. A pattern without wildcards is a valid scenario as the underlying API (GetFiles) supports it - returns a matching file.   This results in a clear and simpler rule: "if you provide a path prefix the path property is a search pattern".  Looking for wildcards in the path, and ignoring selectively ignoring the pathprefix, seems confusing to me and could obfuscate an incorrect client behavior.  

> If output is a directory and there are no wildcards, the directory's path should be used as the pathprefix. The question is whether to figure this out in the node agent or in TES itself.

If the output is a directory, then it returns all files in the directory and the pathprefix is ignored, I am open to revisit this, however, I am failing to understand why this is required as well.

BMurri:
> If the output is a directory, then it returns all files in the directory and the pathprefix is ignored, I am open to revisit this, however, I am failing to understand why this is required as well.

This is required because 4 days ago the entire VM path of each file was appended to the storage directory (instead of the relative path) in the URL, resulting in file paths in the storage container like the following: `Mutect2/e37d88dc-3c69-4131-a4dc-22cdedca8ead/call-SplitIntervals/execution/glob-0fc990c5ca95eebc97c4c204e3e303e1/mnt/batch/tasks/workitems/TES-TSCF7VLI-D4ads_v5-wnvch6bw475so4yjp3hrea56pqnqytwp-oylhbm2k/job-1/e37d88dc_de398853f4b7407bb6335b254c58c6b4-1/wd/wd/cromwell-executions/Mutect2/e37d88dc-3c69-4131-a4dc-22cdedca8ead/call-SplitIntervals/execution/glob-0fc990c5ca95eebc97c4c204e3e303e1/`

giventocode:
> If wildcards are not provided, pathprefix should be ignored. 

I don't see why this is necessary. The implementation assumes that if a pathprefix is set, then pattern search is intended. A pattern without wildcards is a valid scenario as the underlying API (GetFiles) supports it - returns a matching file.   This results in a clear and simpler rule: "if you provide a path prefix the path property is a search pattern".  Looking for wildcards in the path, and ignoring selectively ignoring the pathprefix, seems confusing to me and could obfuscate an incorrect client behavior.  

> If output is a directory and there are no wildcards, the directory's path should be used as the pathprefix. The question is whether to figure this out in the node agent or in TES itself.

If the output is a directory, then it returns all files in the directory and the pathprefix is ignored, I am open to revisit this, however, I am failing to understand why this is required as well.

BMurri (approved):

Issue #267, opened 2023-06-27
Support cross-workspace containers when TES runs on Terra


giventocode:
**Overview**
When TES runs on Terra it assumes a single and predefined storage container.  The assumption facilitates the implementation of the SAS token generation via WSM. However, this approach limits Terra functionality such as workspace cloning, and access to data across workspaces.  The goal of this issue is to change the `TerraStorageProvider` to support inputs and outputs that managed by Terra across workspaces the user has permissions to access. 

AC:
 - Must implement SAS token generation via WSM for inputs and outputs assuming the following:
   - Terra provides only URLs as inputs or outputs to the TES task.
   - The storage container name contains the WSM workspace id.
- Must validate that the provided storage account is a LZ storage account.  
  - The storage account is a LZ storage account, with the following pattern lz[0-9a-f]* as a naming convention.

PR #272, opened 2023-06-30
[267] Add cross-workspace support to the Terra storage provider.

giventocode:
Closes: #267

This PR includes:
 -  ` TerraStorageAccessProvider` generates SAS tokens from any URL Azure storage container managed by Terra provided the user has access to it.
   - The implementation in based on the following assumptions:
     - The storage container name contains the workspace id.
     - The storage account is a valid LZ storage account that follows its naming convention.
 - Removed unnecessary logic to handle local paths as well as a "known" folders  as this logic does not apply to Terra
 - Added basic integration test infrastructure for Terra. 
   - These integration tests are disabled.


src/TesApi.Tests/TerraStorageAccessProviderTests.cs
```diff
@@ -49,81 +50,75 @@ public void SetUp()
...
 
-            var result = await terraStorageAccessProvider.MapLocalPathToSasUrlAsync(input + blobPath, System.Threading.CancellationToken.None, true);
+            var result = await terraStorageAccessProvider.MapLocalPathToSasUrlAsync(input + blobPath, CancellationToken.None, true);
 
             Assert.IsNotNull(result);
             Assert.AreEqual($"{expected}?sv={TerraApiStubData.SasToken}", result);
         }
 
         [TestMethod]
-        [DataRow($"{WorkspaceStorageAccountName}/foo")]
```
BMurri, line 100: Since the path must be an https url, these four inputs should throw exceptions also, correct? Isn't this still a valuable test?

BMurri (changes requested):

BMurri (approved):

PR #292, opened 2023-07-16
Create new Github build/test action and update NetworkUtility tests

MattMcL4475:
Add new build/test actions to Github.  Requires approval for all outside collaborators to run.  

Also updates NetworkUtility to only block outbound network packets, and modifies tests to work in the Github action.

Resolves: https://github.com/microsoft/ga4gh-tes/issues/293


src/Tes.Runner.Test/NetworkUtilityTests.cs
```diff
@@ -29,28 +33,38 @@ public async Task BlockAndUnblockAnIpAddressOnLinux()
             using var client = new HttpClient();
             await client.GetStringAsync(uri);
 
-            await utility.BlockIpAddressAsync(ipAddress);
+            await utility.BlockIpAddressAsync(ipAddress, ruleChain);
+            bool isExceptionThrown = false;
 
             try
             {
                 await client.GetStringAsync(uri);
```
giventocode, line 41: Why not use Assert.ThrowsExceptionAsync<Exception> here?

src/Tes.Runner.Test/NetworkUtilityTests.cs
```diff
@@ -10,16 +10,20 @@ namespace Tes.Runner.Test
...
+        [TestCategory("RequiresRoot")]
         public async Task BlockAndUnblockAnIpAddressOnLinux()
         {
             if (!OperatingSystem.IsLinux())
             {
                 // Not implemented; TES only supports Linux VMs
+                Console.WriteLine("Test did not run because host is not Linux.");
+                Assert.Inconclusive();
                 return;
             }
```
giventocode, line 29: I think you can simply remove the catch block here. You are asserting that no exception must thrown which is what tests implicity do - exceptions fail the test cases. To make it clearer you could name test case along these lines: 
GetStringAsync_IsCalledTwice_SucceedsAndNoExceptionIsThrown

giventocode (approved):
LGTM!

jsaun (approved):

PR #300, opened 2023-07-19
wrap each entry rather than wrap all the entries with && separators

### LINKED ISSUE #284: Executors command with 'bash -c' flag failing task execution
**Describe the bug**
Nextflow ga4gh plugin is POSTing a fixed payload of:

```json
{
  "inputs": [
    {
      "url": "az://work/31/be3fa4b54d73c541cf733f43fd109f/.command.sh",
      "path": "/work/.command.sh",
      "type": "FILE"
    },
    {
      "url": "az://work/31/be3fa4b54d73c541cf733f43fd109f/.command.run",
      "path": "/work/.command.run",
      "type": "FILE"
    },
  ...
  ],
  "outputs": [ 
    ...
  ],
  "resources": {
    "cpu_cores": 1,
    "ram_gb": 0.015625
  },
  "executors": [
    {
      "image": "...",
      "command": [
        "/bin/bash",
        "-c",
        "/bin/bash -ue .command.run &> .command.log"
      ],
      "workdir": "/work"
    }
  ]
}
```
Notice the executors.command attribute, where the second argument is the bash '-c' flag
The command is translated into 
`docker run --rm -v $AZ_BATCH_TASK_WORKING_DIR/wd/work:/work --entrypoint= --workdir /work quay.io/nextflow/bash /bin/bash  "-c && /bin/bash -ue .command.run &> .command.log" && \`
and outputting 
**/bin/bash: - : invalid option**
since the -c flag should have been outside of the quotes:
`/bin/bash  -c "/bin/bash -ue .command.run &> .command.log"
`

**Steps to Reproduce**
run Nextflow with the Azure TES implementation

or

POST /task with executors attribute as such:

```json
  "executors": [
    {
      "image": "artifactory.internal.sophiagenetics.com/docker-all/cprogs:3.46",
      "command": [
        "/bin/bash",
        "-c",
        "/bin/bash -ue .command.run &> .command.log"
      ],
      "workdir": "/work"
    }
  ]
```

**Expected behavior**
TES for Azure should support the -c option, and integrate with Nextflow ga4gh plugin

**Screenshots**
image

BMurri:
fixes #284

### DIFF
--- a/src/TesApi.Web/BatchScheduler.cs
+++ b/src/TesApi.Web/BatchScheduler.cs
@@ -1093,7 +1093,7 @@ namespace TesApi.Web
             sb.AppendLinuxLine($"chmod -R o+rwx $AZ_BATCH_TASK_WORKING_DIR/wd && \\");
             sb.AppendLinuxLine($"export TES_TASK_WD=$AZ_BATCH_TASK_WORKING_DIR/wd && \\");
             sb.AppendLinuxLine($"write_ts ExecutorStart && \\");
-            sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} --entrypoint= --workdir / {executor.Image} {executor.Command[0]}  \"{string.Join(" && ", executor.Command.Skip(1))}\" && \\");
+            sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} --entrypoint= --workdir / {executor.Image} {executor.Command[0]} {string.Join(" ", executor.Command.Skip(1).Select(s => $"\"{s}\""))} && \\");
             sb.AppendLinuxLine($"write_ts ExecutorEnd && \\");
             sb.AppendLinuxLine($"write_ts UploadStart && \\");
             sb.AppendLinuxLine($"./{NodeTaskRunnerFilename} upload --file {UploadFilesScriptFileName} && \\");
--- a/src/TesApi.Web/BatchScheduler.cs
+++ b/src/TesApi.Web/BatchScheduler.cs
@@ -1094,7 +1094,7 @@ namespace TesApi.Web
             sb.AppendLinuxLine($"chmod -R o+rwx $AZ_BATCH_TASK_WORKING_DIR/wd && \\");
             sb.AppendLinuxLine($"export TES_TASK_WD=$AZ_BATCH_TASK_WORKING_DIR/wd && \\");
             sb.AppendLinuxLine($"write_ts ExecutorStart && \\");
-            sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} --entrypoint= {workdirOption}{executor.Image} {executor.Command[0]} {string.Join(" ", executor.Command.Skip(1).Select(s => $"\"{s}\""))} && \\");
+            sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} --entrypoint= {workdirOption}{executor.Image} {executor.Command[0]} {string.Join(" ", executor.Command.Skip(1).Select(BashWrapShellArgument))} && \\");
             sb.AppendLinuxLine($"write_ts ExecutorEnd && \\");
             sb.AppendLinuxLine($"write_ts UploadStart && \\");
             sb.AppendLinuxLine($"./{NodeTaskRunnerFilename} upload --file {UploadFilesScriptFileName} && \\");
@@ -1154,15 +1154,16 @@ namespace TesApi.Web
 
             return cloudTask;
 
+            static string BashWrapShellArgument(string argument)
+                => $"'{argument.Replace(@"'", @"'\''")}'";
+
             static FileType ConvertFileType(TesFileType tesFileType)
-            {
-                return tesFileType switch
+                => tesFileType switch
                 {
                     TesFileType.FILEEnum => FileType.File,
                     TesFileType.DIRECTORYEnum => FileType.Directory,
                     _ => throw new ArgumentOutOfRangeException(nameof(tesFileType)),
-                }; ;
-            }
+                };
 
             // Yes, this looks "Windowsy", while all our executors run on Linux. Environment.ExpandEnvironmentVariables requires environment variables to be delimited by '%' no matter the platform.
             static string LocalizeLocalPath(string path)


src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1154,15 +1154,16 @@ private async Task<CloudTask> ConvertTesTaskToBatchTaskAsync(string taskId, TesT
 
             return cloudTask;
 
+            static string BashWrapShellArgument(string argument)
+                => $"'{argument.Replace(@"'", @"'\''")}'";
```
MattMcL4475, line 1158: Should this escape any special characters?

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1094,7 +1094,7 @@ private async Task<CloudTask> ConvertTesTaskToBatchTaskAsync(string taskId, TesT
             sb.AppendLinuxLine($"chmod -R o+rwx $AZ_BATCH_TASK_WORKING_DIR/wd && \\");
             sb.AppendLinuxLine($"export TES_TASK_WD=$AZ_BATCH_TASK_WORKING_DIR/wd && \\");
             sb.AppendLinuxLine($"write_ts ExecutorStart && \\");
-            sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} --entrypoint= {workdirOption}{executor.Image} {executor.Command[0]}  \"{string.Join(" && ", executor.Command.Skip(1))}\" && \\");
+            sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} --entrypoint= {workdirOption}{executor.Image} {executor.Command[0]} {string.Join(" ", executor.Command.Skip(1).Select(BashWrapShellArgument))} && \\");
```
MattMcL4475, line 1097: Seems like probably should sanitize executor.Command[] to guard against bash injection, unless already done upstream.

MattMcL4475 (approved):

src/TesApi.Web/BatchScheduler.cs
MattMcL4475:
```
sb.AppendLinuxLine($"docker run --rm {volumeMountsOption} --entrypoint= {workdirOption}{executor.Image} " + 
$"{BashWrapShellArgument(executor.Command[0])} " +
$"{string.Join(" ", executor.Command.Skip(1).Select(BashWrapShellArgument))} && \\");

```

MattMcL4475: _For volumeMountsOption, if it's a path, make sure it's a valid path, doesn't contain .., and is under a directory that the application controls. If it's a volume name, validate that it only contains valid characters for a Docker volume name.
For workdirOption, make sure it's a valid path inside the Docker image and doesn't escape from intended directories (like /app)._

MattMcL4475:
```
static string ValidateVolumeMountsOption(string volumeMountsOption)
{
    // Each part should be a valid absolute Unix path.
    var parts = volumeMountsOption.Split(':');
    if (parts.Length != 2 || parts.Any(part => !IsValidUnixPath(part)))
        throw new ArgumentException("Invalid volume mounts option syntax.");

    return volumeMountsOption;
}

static bool IsValidUnixPath(string path)
{
    // Unix paths must start with a /, can't contain null bytes, disallow the .. syntax,
    // and limit to a safe subset of characters (alphanumeric, ., -, _ and /).
    var pattern = @"^(/[\w\.-]+)+$";
    return Regex.IsMatch(path, pattern) && !path.Contains('\0') && !path.Contains("..");
}

static string ValidateDockerImage(string dockerImage)
{
    // Enforce Docker image naming convention: [hostname:port]/[user]/[repository]:[tag]
    var pattern = @"^(?:([^\/]+)\/)?([^:]+)(?::(.+))?$";
    var match = Regex.Match(dockerImage, pattern);
    if (!match.Success)
        throw new ArgumentException("Invalid docker image syntax.");

    // Further validate each part of the Docker image name.
    var hostAndPort = match.Groups[1].Value;
    var repository = match.Groups[2].Value;
    var tag = match.Groups[3].Value;
    
    if (!string.IsNullOrEmpty(hostAndPort) && !IsValidHostAndPort(hostAndPort))
        throw new ArgumentException("Invalid Docker host:port syntax.");
    if (string.IsNullOrEmpty(repository) || !IsValidRepository(repository))
        throw new ArgumentException("Invalid Docker repository syntax.");
    if (!string.IsNullOrEmpty(tag) && !IsValidTag(tag))
        throw new ArgumentException("Invalid Docker tag syntax.");

    return dockerImage;
}

static bool IsValidHostAndPort(string hostAndPort)
{
    // Enforce valid host:port syntax.
    var parts = hostAndPort.Split(':');
    if (parts.Length != 2)
        return false;

    // Check that the host part is a valid hostname or IP address
    if (!Regex.IsMatch(parts[0], @"^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$") &&
        !Regex.IsMatch(parts[0], @"^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$"))
        return false;
    
    // Check that the port part is a valid port number
    if (!int.TryParse(parts[1], out var port) || port < 1 || port > 65535)
        return false;

    return true;
}

static bool IsValidRepository(string repository)
{
    // Docker repositories must be at least two characters long and contain at least one alphanumeric character.
    // They may also contain ., _, and -.
    return Regex.IsMatch(repository, @"^[a-z0-9][a-z0-9\._-]*[a-z0-9]$");
}

static bool IsValidTag(string tag)
{
    // Docker tags must be at least one character long, contain at least one alphanumeric character
    // and may contain ., _, and -.
    return Regex.IsMatch(tag, @"^[A-Za-z0-9][A-Za-z0-9\._-]*[A-Za-z0-9]$") && !tag.All(char.IsDigit);
}

```

SharonHart (approved):

src/TesApi.Web/BatchScheduler.cs
BMurri: This actually covers everything needed for the shell escaping. It does prevent using the command-line to obtain environmental variable values, though. But the goal is to get as close to calling the docker api as possible with the passed in values.

src/TesApi.Web/BatchScheduler.cs
BMurri: That is a great idea, but it should be a separate PR from a separate issue. I've opened one: #307

Issue #145, opened 2023-03-21
Implement SAS resolution strategy for ARM


giventocode:
**Overview**
The TES Task Runner provides an extensibility mechanism to enable just-in-time SAS token resolution on the Batch Nodes. The goal of this story is to implement a new SAS resolution strategy for ARM. 

AC:
- Assumes that the MI of the Batch Node has management access to storage account and can mint SAS tokens using the Keys.
- Should consider using user delegated keys if possible.
- Must create SAS token following the principle of least privilege (POLP) and for the minimum duration possible.
  - Duration must be configurable with a default value of 6 hrs. 
- Must implement how the scheduler should decide to use this strategy depending on configuration values and the input/output information.
- Must define how to distinct public endpoints and set these inputs with the strategy `None` when submitting the task via runner.  
- Must add any new options using the IOptions pattern, if required.
 

DoD:
- Code merge into main, with unit and integration tests that demonstrate the functionality.
- Options documentation update if required.

PR #302, opened 2023-07-20
[145] Azure resource manager SAS resolution for inputs and outputs. 

giventocode:
Closes: #145 

This PR includes:
 - A new SAS resolution strategy for TES runner's inputs and outputs: `ArmSasResolutionStrategy`. 
   - Uses a user delegated key to create the SAS URLs for inputs and outputs. 
   - Uses the default credential provider of the SDK which uses MI, if set and configured the required permissions, to get the user delegated key.
 - SAS URLs are created according to the principle of least privilege - downloads are read-only, uploads have read, write and create permissions.

### DIFF
--- a/src/Tes.Runner/Storage/ArmSasResolutionStrategy.cs
+++ b/src/Tes.Runner/Storage/ArmSasResolutionStrategy.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System.Runtime.InteropServices;
-using Azure;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
 using Azure.Storage.Sas;
--- a/src/Tes.Runner/Storage/ArmSasResolutionStrategy.cs
+++ b/src/Tes.Runner/Storage/ArmSasResolutionStrategy.cs
@@ -1,0 +1,115 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
+using Azure.Storage.Sas;
+using Microsoft.Extensions.Logging;
+using Tes.Runner.Transfer;
+
+namespace Tes.Runner.Storage
+{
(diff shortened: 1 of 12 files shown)


src/Tes.Runner/Storage/ArmSasResolutionStrategy.cs
```diff
@@ -0,0 +1,115 @@
...
+using Azure.Storage.Sas;
+using Microsoft.Extensions.Logging;
+using Tes.Runner.Transfer;
+
+namespace Tes.Runner.Storage
+{
+    public class ArmSasResolutionStrategy : ISasResolutionStrategy
+    {
+        const string StorageHostSuffix = ".blob.core.windows.net";
+        const int BlobSasTokenExpirationInHours = 48;
```
MattMcL4475, line 15: should be 7 days?  Which is the max Azure Batch node runtime

BMurri (approved):

src/Tes.Runner/Storage/ArmSasResolutionStrategy.cs
```diff
@@ -0,0 +1,115 @@
...
+                throw;
+            }
+        }
+
+        private UserDelegationKey GetUserDelegationKey(BlobServiceClient blobServiceClient, string storageAccountName)
+        {
+
+            try
+            {
+                semaphoreSlim.Wait();
```
MattMcL4475, line 84: `WaitAsync() ?`

src/Tes.Runner/Storage/ArmSasResolutionStrategy.cs
```diff
@@ -0,0 +1,115 @@
...
+
+                var userDelegationKey = userDelegationKeyDictionary.GetValueOrDefault(storageAccountName);
+
+                if (userDelegationKey is null || userDelegationKey.SignedExpiresOn < DateTimeOffset.UtcNow)
+                {
+                    userDelegationKey = blobServiceClient.GetUserDelegationKey(startsOn: default, expiresOn: DateTimeOffset.UtcNow.AddHours(UserDelegationKeyExpirationInHours));
```
MattMcL4475, line 90: does this result in an HTTP call?  Is there an `async` version if so?

MattMcL4475 (approved):

Issue #330, opened 2023-08-04
TES fails to create a Batch pool in Terra when a start task is specified


giventocode:
When TES attempts to create a Batch pool with start task the request to WSM fails with the following:

JSON parse error: Cannot construct instance of `bio.terra.workspace.generated.model.ApiAzureBatchPoolAutoUserScope`, problem: Unexpected value 'job' for 'ApiAzureBatchPoolAutoUserScope' enum.; nested exception is com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `bio.terra.workspace.generated.model.ApiAzureBatchPoolAutoUserScope`, problem: Unexpected value 'job' for 'ApiAzureBatchPoolAutoUserScope' enum.

PR #331, opened 2023-08-04
[330] Added missing enum values to API request models to create Batch pools via WSM

giventocode:
Closes: #330 

This PR updates requests models that were mapped to incorrect enum values.


BMurri (approved):
WOW!

PR #332, opened 2023-08-05
Use kubectl port-forward instead of ingress for integration test

### LINKED ISSUE #333: Use kubectl port forward to run the validation test task rather than the ingress
The kubernetes ingress with certificate with letsencrypt is unreliable, for deploying many temporary deployments using kubectl port-forward would be more reliable.

jsaun:
Fixes #333

### DIFF
--- a/src/deploy-tes-on-azure/KubernetesManager.cs
+++ b/src/deploy-tes-on-azure/KubernetesManager.cs
@@ -144,6 +144,15 @@ namespace TesDeployer
             await ExecHelmProcessAsync($"install aad-pod-identity aad-pod-identity/aad-pod-identity --namespace kube-system --version {AadPluginVersion} --kubeconfig \"{kubeConfigPath}\"", cToken);
         }
 
+        /// <summary>
+        /// Enable ingress for TES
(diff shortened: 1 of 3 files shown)


src/deploy-tes-on-azure/KubernetesManager.cs
```diff
@@ -567,18 +591,18 @@ void OutputHandler(object sendingProcess, DataReceivedEventArgs outLine)
             process.Start();
             process.BeginOutputReadLine();
             process.BeginErrorReadLine();
-            await process.WaitForExitAsync();
+            await process.WaitForExitAsync(cToken);
```
BMurri, line 594: Consider process.Kill() if cToken's cancelation state has been triggered.

BMurri (approved):

src/deploy-tes-on-azure/KubernetesManager.cs
jsaun: Internally WaitForExitAsync will kill it if the cTaken is triggered.

BMurri: I didn't see that in the sources for it

PR #340, opened 2023-08-08
[145-1] API clients moved to their own project

giventocode:
Related #145 

This PR includes:
 - Terra and Price API clients, and their tests are now in their own projects.
 - Refactored TES Web project to use the new API client project. 
 - Test for startup DI configuration that validates Terra components are created when the options are set.


BMurri (approved):
lgtm

Please wait until after 4.4.0 is tagged before you merge this.

PR #359, opened 2023-08-11
Ensure container SAS is returned in Terra for uploading start-task logs

### LINKED ISSUE #362: SAS token used for start-task logs is incorrect in Terra
**Describe the bug**
In Terra, the SAS provided to upload the start task logs is a blob sas, not a container sas.

**Expected behavior**
A container sas is needed to upload multiple blobs.

**Additional context**
The default storage access provider and the terra storage access provider do not necessarily return the same types of results given the same inputs.

BMurri:
fixes #362

### DIFF
--- a/src/TesApi.Web/BatchScheduler.cs
+++ b/src/TesApi.Web/BatchScheduler.cs
@@ -1182,7 +1182,7 @@ namespace TesApi.Web
             static string AppendPathToUrl(string url, string path)
             {
                 var uri = new UriBuilder(url);
-                uri.Path += $"/{path}";
+                uri.Path = string.Concat(uri.Path.TrimEnd('/'), "/", path.TrimStart('/'));
                 return uri.Uri.ToString();
             }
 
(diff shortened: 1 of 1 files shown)


src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1179,6 +1179,13 @@ await UploadBlobAsync(NodeTaskRunnerStartTaskFileName, SerializeNodeTask(uploadS
 
             return cloudTask;
 
+            static string AppendPathToUrl(string url, string path)
```
MattMcL4475, line 1182: something like this will ensure / is not a problem:

```
static string AppendPathToUrl(string url, string path)
{
    var uri = new UriBuilder(url);
    string existingPath = uri.Path.TrimEnd('/');
    string appendedPath = path.TrimStart('/');
    uri.Path = string.Concat(existingPath, "/", appendedPath);
    return uri.Uri.ToString();
}
```

BMurri: Done

giventocode (approved):

Issue #367, opened 2023-08-22
Remove path_prefix property from the runner and TES model 


giventocode:
**Overview:**

The node task definition has an output field that contains a list of output files or directories. If an output has a `path_prefix` property, the runner uses the value in the `path` as relative path pattern. The runner uploads the matching files to the storage system. The TES server maps this property to the `path_prefix` field in the TES task model (v 1.1). In addition, the TES server uses this functionality to upload log files after the task is completed. 

However, this property is no longer part of the TES task spec in the latest [proposal](https://github.com/ga4gh/task-execution-schemas/pull/195). Therefore, we need to remove it from the node task definition as well and the TES model on the server.

AC:

 - The runner should assume that the path value is an absolute path pattern and upload the matching files if the output type is FILE or not set.
- Modify the TES server to use the new functionality instead of `path_prefix` when creating additional outputs.

BMurri:
It appears that ga4gh/task-execution-schemas#194 isn't [entirely settled](https://github.com/ga4gh/task-execution-schemas/issues/194#issuecomment-1686643353). This should be deferred until the final resolution is determined.

PR #375, opened 2023-08-26
[367] Removes path-prefix property from the runner and TES task model 

giventocode:
Closes: #367 

In this PR:
 - The path-prefix property in the runner and the TES task model was removed.
 - Output paths are treated as search pattern when the type is FILE.

### DIFF
--- a/src/Tes.Runner/Transfer/DefaultFileInfoProvider.cs
+++ b/src/Tes.Runner/Transfer/DefaultFileInfoProvider.cs
@@ -37,18 +37,40 @@ public class DefaultFileInfoProvider : IFileInfoProvider
     }
 
 
-    public string[] GetFilesBySearchPattern(string path, string searchPattern)
+    public List<FileResult> GetFilesBySearchPattern(string searchPath, string searchPattern)
     {
-        logger.LogInformation($"Searching for files in path: {path} with search pattern: {searchPattern}");
+        logger.LogInformation($"Searching for files in the search path: {searchPath} with search pattern: {searchPattern}");
 
-        return Directory.GetFiles(Environment.ExpandEnvironmentVariables(path), Environment.ExpandEnvironmentVariables(searchPattern), SearchOption.AllDirectories);
+        return Directory.GetFiles(Environment.ExpandEnvironmentVariables(searchPath), Environment.ExpandEnvironmentVariables(searchPattern), SearchOption.AllDirectories)
+            .Select(f => new FileResult(f, ToRelativePathToSearchPath(searchPath, f), searchPath))
+            .ToList();
     }
 
-    public string[] GetAllFilesInDirectory(string path)
+    private string ToRelativePathToSearchPath(string searchPath, string absolutePath)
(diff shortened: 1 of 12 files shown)


src/Tes.Runner/Transfer/DefaultFileInfoProvider.cs
```diff
@@ -37,18 +37,65 @@ public bool FileExists(string fileName)
...
+        return absolutePath;
+    }
+
+    public List<FileResult> GetAllFilesInDirectory(string path)
+    {
+        logger.LogInformation($"Getting all files in directory: {path}");
+
+        if (!Directory.Exists(path))
+        {
+            logger.LogWarning($"The directory provided does not exists: {path}. The output will be ignored.");
```
MattMcL4475, line 75: minor typo "exists" -> "exist"

src/Tes.Runner/Transfer/DefaultFileInfoProvider.cs
```diff
@@ -37,18 +37,65 @@ public bool FileExists(string fileName)
...
+        var delimiter = "/";
 
-        return Directory.GetFiles(Environment.ExpandEnvironmentVariables(path), "*", SearchOption.AllDirectories);
+        if (searchPath.Equals("/", StringComparison.OrdinalIgnoreCase))
+        {
+            delimiter = string.Empty;
+        }
+        var prefixToRemove = Path.GetDirectoryName($"{searchPath}{delimiter}{searchPattern.TrimStart('/')}");
+
+        if (prefixToRemove != null && absolutePath.StartsWith(prefixToRemove))
```
MattMcL4475, line 59: `string.IsNullOrEmpty` ?

MattMcL4475 (approved):

src/Tes/Models/TesOutput.cs
```diff
@@ -60,12 +60,6 @@ public TesOutput()
         [DataMember(Name = "type")]
         public TesFileType Type { get; set; }
 
-        /// <summary>
-        /// Gets or Sets Path Prefix
-        /// </summary>
-        [DataMember(Name = "pathPrefix")]
-        public string PathPrefix { get; set; }
```
BMurri, line 67: Since this property is in the published TES 1.1 spec, should we instead check that the value (if it exists) is equal to the start of `Path` up to (but no including) the point of the level of the first wildcard and throw a not-supported error if not identical? Otherwise we just ignore it.

BMurri (approved):
I'd like to run some tests on this before we merge it to main, but I won't hold it up for that.

src/Tes/Models/TesOutput.cs
giventocode: I understand the point about the spec. However, I see little upside to having the property and not supporting it as defined, albeit its ambiguity. It would not make us 1.1 complaint in any case. I think not having it in the model clearly communicates that we don’t support it, even if it makes our 1.1 support formally incomplete.

Issue #384, opened 2023-09-05
Combined URL transformation strategies for inputs and outputs


giventocode:
**Overview**
The runner applies only one SAS generation strategy to the URL in inputs and outputs. However, there are scenarios where applying multiple strategies to the same URL would be beneficial to avoid complex logic in the TES server to determine what strategy to use for a given input/output. The idea is to have two combined strategies, one for CoA and another for Terra the TES server can set depending on its configuration for all inputs and outputs. These strategies must include the scheme conversion, and the SAS token generation at the very least.

**AC:**
 - Strategies must be applied in order and always return a URL.

PR #385, opened 2023-09-05
[384] Combined URL transformation strategies and refactoring

giventocode:
Closes: #384 

In this PR:
 - The interface ISasResolutionStrategy was renamed to IUrlTransformationStrategy to better describe its intended functionality.
-  Combined transformation strategies for Terra and CoA were introduced. These strategies apply the scheme conversion strategy before applying the respective transformation strategies to obtain SAS tokens according to the runtime environment.


src/Tes.Runner.Test/Storage/TerraUrlTransformationStrategyTests.cs
```diff
@@ -53,11 +48,10 @@ private void SetupWsmClientWithAssumingSuccess()
                     Token = StubSasToken,
                     Url = $"{stubTerraBlobUrl}/{StubBlobName}?{StubSasToken}"
                 })
-                .Callback((Guid wksId, Guid resourceId, SasTokenApiParameters sasTokenParams,
-                    CancellationToken cancellationToken) =>
+                .Callback((Guid _, Guid _, SasTokenApiParameters sasTokenParams,
+                    CancellationToken _) =>
```
BMurri, line 52: For these two lines, sometimes the C# compiler recognizes arguments named `_` as "throwaway" and sometimes it doesn't, which can cause unexpected (and otherwise unexplained) future errors (usually having to do with being unable to cast values to [pick type]). Analyzers have all been upgraded to recognize the pattern "_n" where n is [0-9] (or possibly any series of digits?). So, I'd recommend replacing the three `_` with `_1`, `_2`, & `_3`, respectively.

BMurri (approved):
Please strongly consider the change I recommended. It comes from both experience and official documentation.

Issue #151, opened 2023-03-22
Execute Task commands via the Runner


giventocode:
**Overview**
The TES Task runner has the capability to create a docker container and execute the command on the container using the Docker API. The goal of this story is to integrate this capability with the TES scheduler and add required functionality if needed.'

*AC:*

- Must avoid using the docker container instance and instead using the docker API.
- Must handle the ability to download a container image, for both public container registries and CRs requiring authentication.
- Must delete the container after its execution.
- Must refactor the existing code base execute the task commands via the runner.
- Must return logs from the execution.
- Must add options using the IOptions pattern in the scheduler if required. 

*DoD:*
- Code merged into main, with unit tests for the new functionality, integration tests pass.
- Documented new options if required.

BMurri:
Non-public container registry authentication is now required.

Issue #363, opened 2023-08-14
Move all Batch node initialization and clean-up operations to the runner


giventocode:
*Overview*

The BatchScheduler in the TES server generates, uploads, and manages the execution of bash scripts that perform the following operations on the execution node:
- Move the Docker data/image directory to the data disk.
- Clean-up/prune Docker container images.

However, these operations should be performed before the execution of a container command by the TES runner.

AC:
- All operations must be implemented in a new class that executes these operations and encapsulates their details -e.g. ExecutionEnvironmentManager.
- Must move the current network configuration from the DockerExecutor class to the new class.
- Must implement proper observability (logging, without exposing secrets or sensitive information,) of the execution of these operations to facilitate troubleshooting. 
- Operations must be idempotent. 
- Must integrate the new class with the DockerExecutor as a dependency that is called before the execution of task command on a Docker container.
- Should consider using   `AutoRemove = true` in the host configuration instead of docker pruning commands.

ngambani:
@giventocode can we link this issue to the TES runner https://github.com/microsoft/ga4gh-tes/issues/148? Or should this be separate from the runner implementation?

BMurri:
@MattMcL4475 #515 implements `Clean-up/prune Docker container images.` but does not implement `Move the Docker data/image directory to the data disk.`

If we are going to retain the current behavior of having TES generate start tasks for the docker configuration part, then we should state that and reclose this issue. Alternately, we could open a new issue linking back to this one to move the docker configuration task from the TES server to the TES runner. Otherwise, the issue is only half completed.

BMurri:
Fixed in `6.1.0-beta`

Issue #279, opened 2023-07-07
[#148 Implement TES Task Runner]Private container registry handling through the runner


ngambani:
Motivation to prioritize this: vmsizes that are gen-2 capable are prone to failures in various different ways due to the fact that batch does not have any gen-2 approved images for workloads using containers.

* Increasing security by no longer placing credentials in plain view in the azure pool configuration by placing tes-runner in full control of handing all phases of the interactions with the container registries.
* Improving reliability for Gen2 VMSizes when running in Azure Batch by enabling the docker settings to be fixed BEFORE they are used.
* Improving performance by reducing the time required to queue a TesTask into Azure Batch, while also reducing network traffic and Azure management API calls.

Note: the implementation does NOT work with customizing ACR domain names.

jsaun:
Currently we are providing the ACR admin username and password to Batch https://github.com/microsoft/ga4gh-tes/blob/4.6.0/src/TesApi.Web/BatchScheduler.cs#L1470, when we implement with the TesRunner we should use managed identity instead.

BMurri:
One file was missed

Issue #150, opened 2023-03-22
[#148 Implement TES Task Runner] Implement SAS resolution strategy for DRS


giventocode:
**Overview**
The TES Task Runner provides an extensibility mechanism to enable just-in-time SAS token resolution on the Batch Nodes. The goal of this story is to implement a new SAS resolution strategy when DRS URLs are provided. 

**AC:**

- Must implement the current DRS approach as SAS resolution strategy in the runner.
- Should consider removing the dependency with DRS downloader and perform the resolution calls directly.
- Must add options to the runner to enable DRS SAS resolution .
- Must refactor the scheduler to use this strategy when applicable. 

**DoD:**
- Code merge into main, with unit tests that demonstrate the functionality.
- Documented options for the runner and the API if applicable,

BMurri:
Notes: Consider the following WRT the DRS Localizer (which is the `current DRS approach`):

* Due to the reason for #391 & #390 some of the command-line options we are using are no longer valid (specifically `--vault-name` and `--secret-name`. Both should be removed, along with all the configuration that drives them. See #392.
* The DRS Localizer actually allows all of the DRS downloads to be passed via the command-line in just one call (rather than serially calling the localizer repeatedly for each call). That option is named `--manifest-path` and requires one argument, a CSV file with two columns (`Url`, `Path`) and one row per input file. The file should be without headers and should follow [RFC-4180](https://datatracker.ietf.org/doc/html/rfc4180#section-2) formatting.
* The DRS Localizer uses another option (`--identity-client-id`) for authorization (pass the `ClientId` of the azure managed identity) which we should be setting. This means that the container running the localizer needs to not be blocked.
* The DRS Localizer uses the following environment variables we have not been using: `DRS_RESOLVER_NUM_RETRIES`, `DRS_RESOLVER_WAIT_INITIAL_SECONDS`, `DRS_RESOLVER_WAIT_MAXIMUM_SECONDS`, `DRS_RESOLVER_WAIT_MULTIPLIER`, `DRS_RESOLVER_WAIT_RANDOMIZATION_FACTOR`. We should consider whether we should provide any of these, or simply continue to use the defaults.

The following is provided because it is no longer showing up in the source code:

DRS files are identified via this line of code: `task.Inputs.Where(f => f?.Url?.StartsWith("drs://", StringComparison.OrdinalIgnoreCase) == true)`

[...]

PR #404, opened 2023-09-15
[151] Runner executor

giventocode:
Closes: #151 

In this PR:
  -  TES runner performs the execution of commands (executor).
  -  Moves the batch script preparation (creation and uploading) logic outside the BatchScheduler.
     - `TaskExecutionScriptingManager` Manages the creation and uploading of the Batch script and all its dependencies, this is the entry-point class the BatchScheduler uses.
     - `BatchNodeScriptBuilder`. Fluid builder of the script to be executed by Batch. Used by `TaskExecutionScriptingManager`. 
     - `TaskToNodeTaskConverter`  Handles the creation of a NodeTask from a TesTask. Used by `TaskExecutionScriptingManager`. Creation of the NodeTask's inputs and outputs has the following key functionality:
        -  Handles content inputs by creating them as blobs in storage and setting the NodeTasks' inputs to the blobs' URLs.
        - Handles local file paths in inputs when Cromwell is using the local filesystem with a blob FUSE driver.   

     - `NodeTaskBuilder` Fluid builder of NodeTask instances. Contains the logic that sets the URL transformation strategy for inputs/outputs using the configuration settings. Used by `TaskToNodeTaskConverter`.
  - Added new properties to the NodeTask model:
    - `id` and `workflowId`, uniquely identifies the task and its associated workflow id. 
    - Renamed `SasTransformationStrategy` property in `FileOutput` to `TransformationStrategy`
  - Logging entries and formatting changes to facilitate troubleshooting.
  - Added support for local path mapping to URL for Cromwell local directories (running with blob FUSE driver) and configured external storage. accounts
  - I added the `RuntimeOptions.NodeManagedIdentityResourceId` (workflow ID or global ID as fallback) to the node task and changed the runner to use it `ManagedIdentityCredential(new ResourceIdentifier(runtimeOptions.NodeManagedIdentityResourceId));`
If NodeManagedIdentiyResourceId is not set, the runner will use `DefaultAzureCredentials()` as a fallback, which works when identity is assigned and available.
The TES server now uploads the TES task to storage account.
 - Implemented retries in the runner:
    - When downloading docker images
    - When CredentialUnavailableException is thrown while getting the auth token to generate SAS tokens.
 - Known issues:
   -  DRS URIs are no longer supported, depends on #150 
   -  Private container registry functionality is not supported, to be implemented in #279 
   - Docker image clean up is not supported, to be implemented in #363


src/TesApi.Web/BatchNodeScriptBuilder.cs
```diff
@@ -0,0 +1,168 @@
...
+        private const string BatchTaskDirEnvVar = $"${BatchTaskDirEnvVarName}";
+
+        /// <summary>
+        /// Name of the TES runner binary file
+        /// </summary>
+        public const string NodeTaskRunnerFilename = "tRunner";
+        /// <summary>
+        /// Name of the TES runner task definition file
+        /// </summary>
+        public const string NodeRunnerTaskDefinitionFilename = "TesTask.json";
```
MattMcL4475, line 30: Can we change this name?  Confusing because it's not a Tes.Models.TesTask .  Maybe "NodeTask.json"

giventocode: Yes, I am planning to do file renaming as a follow item, will include changing the binary filename as well as this file.

src/TesApi.Web/BatchNodeScriptBuilder.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <param name="setExecutable">Whether the file should be made executable or not</param>
+        /// <returns>The command to execute</returns>
+        public static string CreateWgetDownloadCommand(string urlToDownload, string localFilePathDownloadLocation, bool setExecutable = false)
+        {
+            if (!Uri.TryCreate(urlToDownload, UriKind.Absolute, out var _))
+            {
+                throw new ArgumentException($"Invalid URL: {urlToDownload}", nameof(urlToDownload));
+            }
+
+            string command = $"wget --https-only --timeout=20 --waitretry=1 --tries=9 --retry-connrefused --continue -O {localFilePathDownloadLocation} '{urlToDownload}'";
```
BMurri, line 158: Add `--no-verbose` to wget command-line

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -977,6 +985,142 @@ private ValueTask<bool> HandleTesTaskTransitionAsync(TesTask tesTask, CombinedBa
...
+        {
+            ValidateTesTask(task);
+
+            var additionalInputs = await GetAdditionalInputsAsync(task, cancellationToken);
+
+            // TODO: Confirm if the mapping here is correct. This is a carry over from the previous implementation. 
+            var poolHasContainerConfig = !(isPublic.ExecutorImage && isPublic.DockerInDockerImage && isPublic.CromwellDrsImage);
+            var containerCleanupOptions = new RuntimeContainerCleanupOptions(
+                ExecuteDockerRmi: !poolHasContainerConfig,
+                ExecuteDockerPrune: enableBatchAutopool);
```
BMurri, line 1000: should be `!enableBatchAutopool`. But as discussed, replace both this and the above line with `true`

src/TesApi.Web/Runner/NodeTaskBuilder.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// </summary>
+        /// <param name="nodeTask"></param>
+        public NodeTaskBuilder(NodeTask nodeTask)
+        {
+            this.nodeTask = nodeTask;
+        }
+
+        private void SetDefaultValues()
+        {
+            nodeTask.ContainerWorkDir = "/";
```
BMurri, line 40: If the container's workdir is not explicitly provided in the TesTask, we don't set it in the docker command-line (or API) (we used to set it to "/", but that was changed over a month ago).

src/TesApi.Web/Runner/TaskToNodeTaskConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+            try
+            {
+                var builder = new NodeTaskBuilder();
+
+                //TODO: Revise this assumption (carried over from the current implementation) and consider Single() if in practice only one executor per task is supported.
+                var executor = task.Executors.First();
+
+                builder.WithId(task.Id)
+                    .WithWorkflowId(task.WorkflowId)
+                    .WithContainerCommands(executor.Command.Select(EscapeBashArgument).ToList())
```
BMurri, line 87: `EscapeBashArgument` is required when each part of the `Command` array from the `TesTask` ends up interpreted by `bash`. If we pass the array directly to the Docker API (which we do here), then they should NOT be escaped.

src/TesApi.Web/Runner/TaskToNodeTaskConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+            return uri.GetLeftPart(UriPartial.Path);
+        }
+
+        private static string EscapeBashArgument(string arg)
+        {
+            return $"{arg.Replace(@"'", @"'\''")}";
+        }
+
+
+        private static void MapOutputs(List<TesOutput> outputs, string pathParentDirectory, string containerMountParentDirectory,
```
BMurri, line 258: This needs to convert file schema urls to storage account urls

src/TesApi.Web/Runner/TaskToNodeTaskConverter.cs
```diff
@@ -0,0 +1,308 @@
...
+        {
+            outputs?.ForEach(output =>
+            {
+                builder.WithOutputUsingCombinedTransformationStrategy(
+                    AppendParentDirectoryIfSet(output.Path, pathParentDirectory), output.Url, ToNodeTaskFileType(output.Type),
+                    containerMountParentDirectory);
+            });
+        }
+
+        private static void MapInputs(List<TesInput> inputs, string pathParentDirectory, string containerMountParentDirectory,
```
BMurri, line 269: Same as MapOutputs

BMurri (changes requested):

src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
```diff
@@ -166,6 +167,24 @@ private static string NormalizedBlobPath(string blobPath)
...
+            var blobPathWithPrefix = $"{task.Id}{normalizedBlobPath}";
+            if (task.Resources?.ContainsBackendParameterValue(TesResources.SupportedBackendParameters
+                    .internal_path_prefix) == true)
+            {
+                blobPathWithPrefix =
+                    $"{task.Resources.GetBackendParameterValue(TesResources.SupportedBackendParameters.internal_path_prefix).Trim('/')}{normalizedBlobPath}";
+            }
+
+            //passing the resulting string through the builder to ensure that the path is properly encoded and valid
+            var builder = new BlobUriBuilder(new Uri($"https://{defaultStorageAccountName}.blob.core.windows.net/{blobPathWithPrefix.TrimStart('/')}"));
```
MattMcL4475, line 183: is this missing the container name?

src/TesApi.Web/Startup.cs
```diff
@@ -107,6 +108,9 @@ public void ConfigureServices(IServiceCollection services)
                     .AddSingleton<ConfigurationUtils>()
                     .AddSingleton<IAllowedVmSizesService, AllowedVmSizesService>()
                     .AddSingleton<TokenCredential>(s => new DefaultAzureCredential())
+                    .AddSingleton<TaskToNodeTaskConverter>()
+                    .AddSingleton<TaskExecutionScriptingManager>()
+                    .AddTransient<BatchNodeScriptBuilder>()
```
MattMcL4475, line 113: Isn't this a utility class?  Is it really appropriate for dependency injection?

src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
giventocode: Yes, good catch. This method is not used. I will clean this up.

src/TesApi.Web/Startup.cs
giventocode: hmm, not sure I agree. These are dependencies that are required to be injected and their creation requires options and other dependencies from the app context.  I think DI is doing what is supposed to be do.

src/Tes.Runner/Docker/DockerExecutor.cs
```diff
@@ -61,16 +61,17 @@ private async Task<MultiplexedStream> StreamStdOutAndErrorAsync(string container
...
-        private async Task<CreateContainerResponse> CreateContainerAsync(string imageName, List<string> commandsToExecute, List<string>? volumeBindings)
+        private async Task<CreateContainerResponse> CreateContainerAsync(string imageName, string imageTag,
+            List<string> commandsToExecute, List<string>? volumeBindings, string? workingDir)
         {
             var createResponse = await dockerClient.Containers.CreateContainerAsync(
                 new CreateContainerParameters
                 {
-                    Image = imageName,
+                    Image = ToImageNameWithTag(imageName, imageTag),
                     Entrypoint = commandsToExecute,
```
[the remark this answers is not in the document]
MattMcL4475: we set it to an empty string

MattMcL4475 (approved):

src/CommonUtilities/Models/NodeTask.cs
```diff
@@ -42,7 +42,7 @@ public class RuntimeOptions
     {
         public TerraRuntimeOptions? Terra { get; set; }
 
-        public DockerCleanUpOptions DockerCleanUp { get; set; }
+        public string? NodeManagedIdentityResourceId { get; set; }
     }
 
     public class DockerCleanUpOptions
```
BMurri, line 48: If this is no longer used, please remove.

src/Tes.Runner/Authentication/CredentialsManager.cs
```diff
@@ -0,0 +1,64 @@
...
+            var duration = TimeSpan.FromSeconds(Math.Pow(ExponentialBackOffExponent, attempt));
+            logger.LogInformation($"Waiting {duration} before retrying");
+            return duration;
+        }
+
+        public TokenCredential GetTokenCredential(RuntimeOptions runtimeOptions)
+        {
+            return retryPolicy.Execute(() => GetTokenCredentialImpl(runtimeOptions));
+        }
+        private TokenCredential GetTokenCredentialImpl(RuntimeOptions runtimeOptions)
```
BMurri, line 42: Need a blank line here

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -169,8 +154,7 @@ public partial class BatchScheduler : IBatchScheduler
                 _batchPoolFactory = poolFactory;
                 batchPrefix = batchSchedulingOptions.Value.Prefix;
                 logger.LogInformation("BatchPrefix: {BatchPrefix}", batchPrefix);
-                taskRunScriptContent = File.ReadAllLines(Path.Combine(AppContext.BaseDirectory, "scripts/task-run.sh"));
-                taskCleanupScriptContent = File.ReadAllLines(Path.Combine(AppContext.BaseDirectory, "scripts/clean-executor.sh"));
+                File.ReadAllLines(Path.Combine(AppContext.BaseDirectory, "scripts/task-run.sh"));
```
BMurri, line 157: If this isn't storing the contents of the file anywhere, this line can go away. If that is the case, please also delete the two files in the `scripts` directory.

src/TesApi.Web/Runner/TaskToNodeTaskConverter.cs
```diff
@@ -0,0 +1,417 @@
...
+            {
+                return default;
+            }
+
+            logger.LogInformation($"The input is content. Uploading its content to the internal storage location. Input path:{input.Path}");
+
+            return await UploadContentAndCreateTesInputAsync(tesTask, input.Path, input.Content, cancellationToken);
+        }
+
+        private static string EscapeBashArgument(string arg)
```
BMurri, line 352: Are you dumping the commands from the TesTask into a script in a single string or are you passing them to the Docker API? If the latter, we should NOT be shell escaping them. The Docker API does not turn them into a single string command-line, it passes them to the process's main() function directly.

BMurri (approved):
Mostly nitpicks and cleanup. Please open issues for any you don't do now.

src/CommonUtilities/Models/NodeTask.cs
BMurri: Should have been line 48

PR #406, opened 2023-09-17
Make WriterWorkerProc a simple async Task and only delay 50ms

### LINKED ISSUE #405: Unexpected slowness: POST TaskServiceApi/CreateTask is taking 512ms on average
Research why it's taking this long and if there are obvious improvements that can be made.  

image

Note: PostgreSQL is not busy:
image

vs. GET:

image

MattMcL4475:
Currently, posting a TES task takes on average 500ms due to the WriterWorkerProc wait time.  This PR reduces the wait to 50ms (which should result in an average of 25ms), and also rewrites it to be async and removes BackgroundWorker.

Resolves #405

### DIFF
--- a/src/Tes/Repository/PostgreSqlCachingRepository.cs
+++ b/src/Tes/Repository/PostgreSqlCachingRepository.cs
@@ -42,7 +42,7 @@ namespace Tes.Repository
         {
             _logger = logger;
             _cache = cache;
-            _writerWorkerTask = Task.Run(WriterWorkerAsync);
+            _writerWorkerTask = Task.Run(() => WriterWorkerAsync(_writerWorkerCancellationTokenSource.Token));
         }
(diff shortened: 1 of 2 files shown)


src/Tes/Repository/PostgreSqlCachingRepository.cs
```diff
@@ -135,7 +152,7 @@ Task<T> RemoveUpdatingItem(Task<T> task)
         /// <summary>
         /// Continuously writes items to the database
         /// </summary>
-        private async Task WriterWorkerAsync(CancellationToken cancellationToken)
+        private async ValueTask WriterWorkerAsync(CancellationToken cancellationToken)
```
MattMcL4475, line 155: I'm not sure of the implications of this, if any

PR #410, opened 2023-09-19
Reset batch's autoscale when batch fails to run autoscale in a long time

### LINKED ISSUE #409: Mitigate batch error where old allocation errors cause batch to fail to evaluate the autoscale formula
**Describe the bug**
Occasionally Azure Batch will fail to run the collections needed for autoscale to function after a quota failure. When that happens, the pool becomes "stuck" and never resizes. The current code does not recognize this state and thus does not mitigate it.

**Steps to Reproduce**
Run workflows where the batch account's quota is exceeded. Then, after some time, run more tasks. You MAY see this. If not, keep trying. It'll eventually show up.

**Expected behavior**
Either the failure in batch will not happen (ideal) or our code will recognize this and "kick-restart" batch's use of autoscale.

BMurri:
fixes #409

### DIFF
--- a/src/TesApi.Tests/BatchPoolTests.cs
+++ b/src/TesApi.Tests/BatchPoolTests.cs
@@ -97,6 +97,24 @@ namespace TesApi.Tests
             }
         }
 
+        [TestMethod]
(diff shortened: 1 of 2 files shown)


MattMcL4475 (approved):
LGTM

PR #411, opened 2023-09-21
Reduce aks node pool size to 1

jsaun:


BMurri:
Would require same kinds of changes as microsoft/CromwellOnAzure#704

PR #420, opened 2023-09-25
Overshoot compute node targets by rounding up to nearest 10

BMurri:


src/TesApi.Web/BatchPool.cs
```diff
@@ -221,7 +223,9 @@ public static string AutoPoolFormula(bool preemptable, int initialTarget)
                 "span = TimeInterval_Second * 90;",
                 "startup = TimeInterval_Minute * 2;",
                 "ratio = 10;",
```
MattMcL4475, line 225: `{RoundUpOvershoot}` should be used here so it's defined once

src/TesApi.Web/BatchPool.cs
```diff
@@ -31,6 +31,8 @@ public sealed partial class BatchPool
 
         private const int MaxComputeNodesToRemoveAtOnce = 100; // https://learn.microsoft.com/en-us/rest/api/batchservice/pool/remove-nodes?tabs=HTTP#request-body nodeList description
 
+        private const int RoundUpOvershoot = 10;
```
MattMcL4475, line 34: Let's set this in the YAML config so we can adjust it at runtime

MattMcL4475 (changes requested):

Issue #432, opened 2023-09-30
Edge case whereby the DB schema is not updated and cache not warmed before enabling REST API


MattMcL4475:
Edge case could occur whereby the DB schema is not updated and cache not warmed before enabling REST API, leading to corrupt state of tes tasks.

MattMcL4475:
This edge case actually cannot happen, since `.AddSingleton<TesTaskPostgreSqlRepository>()` guarantees that `TesTaskPostgreSqlRepository`'s constructor completes.

PR #433, opened 2023-09-30
Ensure database migration and cache warming complete before enabling API

MattMcL4475:
- Fixes potential edge case.  Ensures DB schema is updated and cache warmed before enabling REST API (#432)
- Remove double instantiation of Batch Scheduler

Note: `ActivatorUtilities.CreateInstance<TesTaskPostgreSqlRepository>(sp)` forces the completion of the constructor for `TesTaskPostgreSqlRepository` and also the injection of its constructor dependencies.  This ensures that the following completes:
```csharp
dbContext.Database.MigrateAsync().Wait();
WarmCacheAsync(CancellationToken.None).Wait();
```

### DIFF
--- a/src/TesApi.Web/Startup.cs
+++ b/src/TesApi.Web/Startup.cs
@@ -75,7 +75,7 @@ namespace TesApi.Web
 
                     .AddMemoryCache(o => o.ExpirationScanFrequency = TimeSpan.FromHours(12))
                     .AddSingleton<ICache<TesTaskDatabaseItem>, TesRepositoryCache<TesTaskDatabaseItem>>()
-                    .AddSingleton<TesTaskPostgreSqlRepository>()
+                    .AddSingleton(sp => ActivatorUtilities.CreateInstance<TesTaskPostgreSqlRepository>(sp))
                     .AddSingleton<AzureProxy>()
                     .AddTransient<BatchPool>()
                     .AddSingleton<IBatchPoolFactory, BatchPoolFactory>()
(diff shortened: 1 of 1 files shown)


src/TesApi.Web/Startup.cs
```diff
@@ -89,7 +89,6 @@ public void ConfigureServices(IServiceCollection services)
                         })
                     .Services
 
-                    .AddSingleton<IBatchScheduler, BatchScheduler>()
```
MattMcL4475, line 92: @BMurri  is this the right one to remove of the two?

src/TesApi.Web/Startup.cs
```diff
@@ -75,7 +75,7 @@ public void ConfigureServices(IServiceCollection services)
 
                     .AddMemoryCache(o => o.ExpirationScanFrequency = TimeSpan.FromHours(12))
                     .AddSingleton<ICache<TesTaskDatabaseItem>, TesRepositoryCache<TesTaskDatabaseItem>>()
-                    .AddSingleton<TesTaskPostgreSqlRepository>()
+                    .AddSingleton(sp => ActivatorUtilities.CreateInstance<TesTaskPostgreSqlRepository>(sp))
```
BMurri, line 78: This change is redundant (the other syntax does exactly this, and it was easier to read)

BMurri (approved):
Duplicate Singleton registrations replace the former leaving the later but it is incorrect to have both (I looked and it was introduced in a merge).

I'd suggest reverting the repository change. It's less readable and does the exact same thing.

Issue #424, opened 2023-09-27
Reduce time it takes to notice a task is complete 


MattMcL4475:
1) TES will no longer ask Batch for complete information on each task each time. Instead,
2) The controller will set each accepted task to `QUEUED` as it does today. That'll mean it is not assigned to a pool, yet.
3) BatchScheduler will set QUEUED tasks to `INITIALIZING` when assigning them to a pool. CloudTasks will be set to immediately expire when complete.
4) Scheduler/BatchScheduler will check the batch account blob for the current status of each task. When it's complete, it'll be thus processed.
5) A per-pool query for failed tasks will be performed for failed tasks that will be thus processed.
6) All the rest of the "error handling"/"hand holding" will be removed, including support for AutoPools.

In the future, #4 can be updated to cover other states, like "running", etc.

Task information beyond state will only be obtained when needed, not every loop. This will greatly diminish the number of calls to Batch, and greatly lessen the work TES is doing

BMurri:
Fixed in `6.1.0-beta`

PR #442, opened 2023-10-04
Simplify task management

### LINKED ISSUE #627: Minimize the risk of expired SAS tokens AND simplify node management
**Problem:**
For every work task, two things sited in azure blob storage MUST be on the node: the task runner & the task runner's task json. Those things currently must be tagged with a SAS token because they cannot be downloaded without it. Any start task that needs any resource from blob storage suffers from the same issue.

Any task added to a job cannot have its command-line changed (e.g. to update a SAS token) without first terminating and then deleting the task from the job and replacing it with a new one (which will end up at the end of the line). This is a problem when running at scale, because it is very conceivable (and actually has happened) that the token expires before the task finally starts running.

Any start tasks that must download anything requiring a SAS token have it worse, because the start task is generated at pool creation (and thus becomes a long-lived entity). In Terra today, SAS tokens live shorter lives than pools do (and a pools "lifetime" setting is the limit for new task additions to the pool's job, NOT new task STARTs). Start tasks can be updated, but that appears to require either a different batch client (with the C# library) than the one we are currently using, or a different approach to how we call the batch data-plane APIs than what we are currently doing.

**Solution:**
1. As proposed in #520, load the node task runner via the startup task. As an expansion of #363, perform all start-task related work via that runner. Further, use that runner for all tasks scheduled/run on that node.
2. Alter the runner such that it accepts from its command-line and/or environment variables all information needed to be able to generate a SAS token and download the task JSON (thus, eliminating the need for the TES server to supply any SAS token in the task command-line/task script file).

Do nothing knowing that these issues will continue to be issues, especially as environments ask for shorter SAS token lifetimes as time goes on.

**Sub Tasks**
- #629 
- #628 
- #630 
- #631 

**Code dependencies**
* `CoA, for new and/or existing deployments`? No
* `TES standalone, for new and/or existing deployments`? No
* `Terra, for new and/or existing deployments`? No
* `Build pipeline`? No
* `Integration tests`? No

**Additional context**
Completing this feature will enable easier implementation and/or largely or fully complete the following issues:
* #363 
* #454 
* #520 
* microsoft/CromwellOnAzure#102 
* microsoft/CromwellOnAzure#206 
* microsoft/CromwellOnAzure#243 
* microsoft/CromwellOnAzure#368 
* microsoft/CromwellOnAzure#402 
* microsoft/CromwellOnAzure#408 
* microsoft/CromwellOnAzure#410 
* microsoft/CromwellOnAzure#671 
*  A reimagination of the issues that underpin microsoft/CromwellOnAzure#402

### LINKED ISSUE #631: [#627 expired SAS tokens] Call the runner from the start task
**Problem:**
The current method of managing "global" start tasks set in configuration, compute node image-based start-task commands used to configure the docker daemon service, and future per-vmfamily start-task activities (including configurable/installable ones as [envisioned for Illumina DRAGEN](https://github.com/microsoft/CromwellOnAzure/issues/317) support) is really brittle.

**Solution:**
The runner should be able to run multiple scripts. Consider whether to extend the current task JSON or to instead create a new schema for start tasks. Then, supply the start-task's JSON's (in whatever form it ends up in) URL in the start task command-line along with the bash/shell commands to download the runner's binary and `chmod` it. Consider building into the runner start-task tasks that we know will likely be permanent (such as configuring the docker daemon). Also consider uploading the start task's trace files (`stderr.txt` & `stdout.txt` to an appropriate location in the `tes-internals` container space.

BMurri:
Addresses microsoft/CromwellOnAzure#404
Addresses microsoft/CromwellOnAzure#523
Addresses #424
Required by microsoft/CromwellOnAzure#709
fixes #631
fixes #627


src/deploy-tes-on-azure/Deployer.cs
```diff
@@ -327,12 +329,21 @@ public async Task<int> DeployAsync()
...
+
                             if (string.IsNullOrWhiteSpace(settings["BatchNodesSubnetId"]))
                             {
                                 settings["BatchNodesSubnetId"] = await UpdateVnetWithBatchSubnet();
                             }
                         }
 
+                        if (installedVersion is null || installedVersion < new Version(4, 7))
+                        {
+                            // Storage account now requires Storage Blob Data Owner
```
MattMcL4475, line 343: `Storage Blob Data Contributor` has "Read, write, and delete Azure Storage containers and blobs".  Just wanted to double check that Owner is really required

https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles

src/TesApi.Web/TesEventMessage.cs
```diff
@@ -0,0 +1,178 @@
...
+            await _azureProxy.SetBlobTags(
+                _uri,
+                Tags.Append(new(ProcessedTag, DateTime.UtcNow.ToString("O")))
+                    .ToDictionary(pair => pair.Key, pair => pair.Value),
+                cancellationToken);
+        }
+
+        private /*static*/ AzureBatchTaskState GetCompletedBatchState(Tes.Runner.Events.EventMessage task)
+        {
+            return (task.Name ?? Event) switch
```
giventocode, line 128: A better way to verify the event is to look at its statusMessage rather than the error message.

src/TesApi.Web/TesEventMessage.cs
```diff
@@ -0,0 +1,178 @@
...
+using Microsoft.Extensions.Logging;
+using TesApi.Web.Storage;
+
+namespace TesApi.Web
+{
+    /// <summary>
+    /// Represents the events sent by the node task runner.
+    /// </summary>
+    /// <remarks>This should be transient in DI.</remarks>
+    public class TesEventMessage
```
giventocode, line 19: The name does not reflect the purpose of this class. It acts as a message processor, not a data object or entity. A better name could be: RunnerEventsProcessor, RunnerEventsHandler or RunnerEventsConsumer.

src/TesApi.Web/TesEventMessage.cs
```diff
@@ -0,0 +1,178 @@
...
+    {
+        static TesEventMessage() => Tes.Utilities.NewtonsoftJsonSafeInit.SetDefaultSettings();
```
giventocode, line 21: I think we want to avoid new usages of NewtonSoft. The runner library has a method for deserialization. Maybe you can use it, make it public if needed. We should use only one method in the code base for deserialization and serialization of the class.

src/TesApi.Web/TesEventMessage.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        /// <summary>
+        /// Blob tag used to record event processing.
+        /// </summary>
+        public const string ProcessedTag = "processed";
+
+        private readonly IStorageAccessProvider _storageAccessProvider;
```
giventocode, line 28: This notation is not used for the most part in the code base. We should avoid it for consistency's sake, even if it is better than the current style/pattern.

src/TesApi.Web/TesEventMessage.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// <summary>
+        /// Constructor of <see cref="TesEventMessage"/>.
+        /// </summary>
+        /// <param name="azureProxy"></param>
+        /// <param name="logger"></param>
+        /// <param name="storageAccessProvider"></param>
+        /// <param name="blobAbsoluteUri"></param>
+        /// <param name="tags"></param>
+        /// <param name="event"></param>
+        public TesEventMessage(IAzureProxy azureProxy, ILogger<TesEventMessage> logger, IStorageAccessProvider storageAccessProvider, Uri blobAbsoluteUri, IDictionary<string, string> tags, string @event)
```
giventocode, line 52: Logger typically is the last argument in the constructor....

src/TesApi.Web/TesEventMessage.cs
```diff
@@ -0,0 +1,178 @@
...
+            Tags = tags.AsReadOnly();
+            Event = @event;
+        }
+
+        /// <summary>
+        /// Gets the details of this event message.
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task<(string Id, AzureBatchTaskState State)> GetMessageBatchStateAsync(CancellationToken cancellationToken)
```
giventocode, line 90: Can we return a type here instead of multiple values?

src/TesApi.Web/TesEventMessage.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        /// <summary>
+        /// Marks this event message processed.
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task MarkMessageProcessed(CancellationToken cancellationToken)
+        {
+            await _azureProxy.SetBlobTags(
+                _uri,
```
giventocode, line 120: What is the lifecycle of this URI?  Does it have the SAS token and if so, is there a risk it could expire?

src/TesApi.Web/TesEventMessage.cs
```diff
@@ -0,0 +1,178 @@
...
+        public BatchTesEventMessageFactory(IServiceProvider serviceProvider) => _serviceProvider = serviceProvider;
+
+        /// <summary>
+        /// Creates a new <see cref="TesEventMessage"/>.
+        /// </summary>
+        /// <param name="blobAbsoluteUri"></param>
+        /// <param name="tags"></param>
+        /// <param name="event"></param>
+        /// <returns></returns>
+        public TesEventMessage CreateNew(Uri blobAbsoluteUri, IDictionary<string, string> tags, string @event)
```
giventocode, line 177: This class exists because the tags, event, and URI are part of the constructor, and they change per instance. Instead, I would pass them to the method that uses them, simplify the constructor, and register the type at start up directly in DI.

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -181,157 +177,141 @@ public partial class BatchScheduler : IBatchScheduler
...
-                new TesTaskStateTransition(tesTaskIsQueuedInitializingOrRunning, BatchTaskState.NodeFailedDuringStartupOrExecution, "Please open an issue. There should have been an error reported here.", DeleteBatchJobAndSetTaskExecutorErrorAsync),
-                new TesTaskStateTransition(tesTaskIsQueuedInitializingOrRunning, BatchTaskState.NodeUnusable, "Please open an issue. There should have been an error reported here.", DeleteBatchJobAndSetTaskExecutorErrorAsync),
-                new TesTaskStateTransition(tesTaskIsInitializingOrRunning, BatchTaskState.JobNotFound, BatchTaskState.JobNotFound.ToString(), SetTaskSystemError),
-                new TesTaskStateTransition(tesTaskIsInitializingOrRunning, BatchTaskState.MissingBatchTask, BatchTaskState.MissingBatchTask.ToString(), DeleteBatchJobAndSetTaskSystemErrorAsync),
-                new TesTaskStateTransition(tesTaskIsInitializingOrRunning, BatchTaskState.NodePreempted, alternateSystemLogItem: null, HandlePreemptedNodeAsync)
-            };
+                new TesTaskStateTransition(tesTaskDeletionReady, batchTaskState: null, alternateSystemLogItem: null, (tesTask, _, ct) => DeleteCancelledTaskAsync(tesTask, ct)),
```
giventocode, line 266: I think this constructor is way too long and complex, and this class does too much.  Could the state transitions be refactored into their own class - i.e. TaskStateTransitionManager?

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -491,125 +456,171 @@ private static string StandardizeStartTaskPath(string startTaskPath, string defa
...
 
-        /// <summary>
-        /// Adds a new Azure Batch pool/job/task for the given <see cref="TesTask"/>
-        /// </summary>
-        /// <param name="tesTask">The <see cref="TesTask"/> to schedule on Azure Batch</param>
-        /// <param name="cancellationToken">A <see cref="CancellationToken"/> for controlling the lifetime of the asynchronous operation.</param>
-        /// <returns>A task to await</returns>
-        private async Task AddBatchTaskAsync(TesTask tesTask, CancellationToken cancellationToken)
+        /// <inheritdoc/>
+        public async IAsyncEnumerable<(TesTask TesTask, Task<bool> IsModifiedAsync)> ProcessQueuedTesTasksAsync(TesTask[] tesTasks, [System.Runtime.CompilerServices.EnumeratorCancellation] CancellationToken cancellationToken)
```
giventocode, line 460: Why use a new type instead of multiple return values? A new type can make the method's purpose more clear and allow for more flexibility in adding or removing properties as needed.

src/Tes.ApiClients/RetryHandler.cs
```diff
@@ -132,3 +166,25 @@ public virtual async Task<HttpResponseMessage> ExecuteHttpRequestWithRetryAsync(
         return await asyncHttpRetryPolicy.ExecuteAsync(action, cancellationToken);
     }
 }
+
+public static class RetryHandlerExtensions
```
giventocode, line 170: What is the purpose of these methods, I don't see them used anywhere? Perhaps, I missed it.

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1724,22 +1326,67 @@ static bool TryGetValueAsDouble(Dictionary<string, string> dict, string key, out
                 .Select(line => { var parts = line.Split(fieldDelimiter); return new KeyValuePair<string, string>(parts[0], parts[1]); })
                 .ToDictionary(kv => kv.Key, kv => kv.Value);
 
+
+        /// <inheritdoc/>
+        public async IAsyncEnumerable<TesEventMessage> GetEventMessagesAsync([System.Runtime.CompilerServices.EnumeratorCancellation] CancellationToken cancellationToken, string @event)
```
giventocode, line 1331: A more suitable place for this method is the TestEventMessage (RunnerEventsConsumer) class, not this class. There is no need to create a new instance of BatchScheduler to get messages.

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1724,22 +1326,67 @@ static bool TryGetValueAsDouble(Dictionary<string, string> dict, string key, out
...
+            {
+                prefix += @event + "/";
+                //tags.Add("event-name", @event);
+            }
+
+            var tesInternalSegments = StorageAccountUrlSegments.Create(storageAccessProvider.GetInternalTesBlobUrlWithoutSasToken(string.Empty));
+            var eventsStartIndex = (string.IsNullOrEmpty(tesInternalSegments.BlobName) ? string.Empty : (tesInternalSegments.BlobName + "/")).Length;
+            var eventsEndIndex = eventsStartIndex + eventsFolderName.Length + 1;
+
+            //await foreach (var blobItem in azureProxy.ListBlobsWithTagsAsync(new(await storageAccessProvider.GetInternalTesBlobUrlAsync(string.Empty, cancellationToken, needsFind: true)), prefix, tags, cancellationToken).WithCancellation(cancellationToken))
```
giventocode, line 1348: Remove commented out code.

giventocode (changes requested):

BMurri:
@giventocode I have yet to do a LOT of cleanup. This code right now is littered with the debris of a lot of testing and various routes taken on portions of the implementation. Once I have the end-to-end running, the next step is to clean it all up and restore/enlarge the testing. That's a huge part of the reason this PR is still in `draft`

giventocode:
> @giventocode I have yet to do a LOT of cleanup. This code right now is littered with the debris of a lot of testing and various routes taken on portions of the implementation. Once I have the end-to-end running, the next step is to clean it all up and restore/enlarge the testing. That's a huge part of the reason this PR is still in `draft`

Sounds good. Let me know when you want me to another pass at it.

src/Tes.Runner/Events/EventsPublisher.cs
```diff
@@ -11,8 +11,8 @@ namespace Tes.Runner.Events;
 
 public class EventsPublisher : IAsyncDisposable
 {
-    const string EventVersion = "1.0";
-    const string EventDataVersion = "1.0";
+    public const string EventVersion = "1.0";
```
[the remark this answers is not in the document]
BMurri: In `src/TesApi.Web/Events/NodeEventMessage.cs`

src/TesApi.Web/BatchScheduler.cs
BMurri: We don't create a new instance of `BatchScheduler`. We use the singleton from `Startup` that DI gives us.

src/TesApi.Web/BatchScheduler.cs
BMurri: They could be, and in fact I opened an issue about refactoring this class into multiple classes a long time ago, but this is how this class is/has been organized since the beginning. That refactor belongs in its own PR IMHO

src/TesApi.Web/TesEventMessage.cs
BMurri: From a single call, all non-processed event blobs are gathered, SAS tokens are obtained, and all events are processed (which consists of updating `TesTask`s in the database), in parallel as much as possible. The only way that the token could expire is essentially if the tes task database were down so long and the db retry calls lasted long enough for the SAS to expire (not likely).

src/TesApi.Web/TesEventMessage.cs
BMurri: Working on that...

src/TesApi.Web/OrchestrateOnBatchSchedulerServiceBase.cs
```diff
@@ -0,0 +1,273 @@
...
+        {
+            return GetType().Name + " " + message;
+        }
+
+        /// <inheritdoc />
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA2254:Template should be a static expression", Justification = "Used to provide service's name in log message.")]
+        public override Task StopAsync(CancellationToken cancellationToken)
+        {
+            logger.LogInformation(MarkLogMessage("stopping..."));
+            return base.StopAsync(cancellationToken);
```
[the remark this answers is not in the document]
BMurri: The base class takes care of that. This exists simply to log

src/TesApi.Web/Management/BatchQuotaVerifier.cs
```diff
@@ -106,30 +107,40 @@ public async Task CheckBatchAccountQuotasAsync(VirtualMachineInformation virtual
...
-                throw new AzureBatchQuotaMaxedOutException($"Not enough core quota remaining to schedule task requiring {workflowCoresRequirement} {(isDedicated ? "dedicated" : "low priority")} cores. There are {batchUtilization.TotalCoresInUse} cores in use out of {batchVmFamilyBatchQuotas.TotalCoreQuota}.");
-            }
+            throw new AzureBatchQuotaMaxedOutException($"Not enough core quota remaining to schedule task requiring {workflowCoresRequirement} {(isDedicated ? "dedicated" : "low priority")} cores. There are {batchUtilization.TotalCoresInUse} cores in use out of {batchVmFamilyBatchQuotas.TotalCoreQuota}.");
+        }
+    }
 
-            if (isDedicatedAndPerVmFamilyCoreQuotaEnforced && batchUtilization.DedicatedCoresInUseInRequestedVmFamily + workflowCoresRequirement > batchVmFamilyBatchQuotas.VmFamilyQuota)
-            {
+    /// <inheritdoc cref="IBatchQuotaProvider"/>
+    public async Task<(int exceeded, Exception exception)> CheckBatchAccountPoolOrJobQuotasAsync(int required, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
BMurri: It's used in `BatchScheduler` to extend the quota checks to cover the case where multiple pools are being added in one call from the Scheduler, since we now add tasks in parallel in groups instead of serially and the first round of quota checks assume that only one pool/job are added at a time.

Operationally, two kinds of values are returned: exceeded > 0 and exception is not null, where exceeded is the number of new pools that cannot be created because of quota, or exceeded == 0 and exception is null, where all the requested pools can be created.

src/TesApi.Web/Management/BatchQuotaVerifier.cs
```diff
@@ -106,30 +107,40 @@ public async Task CheckBatchAccountQuotasAsync(VirtualMachineInformation virtual
...
         if (needPoolOrJobQuotaCheck)
         {
-            if (batchUtilization.ActiveJobsCount + 1 > batchVmFamilyBatchQuotas.ActiveJobAndJobScheduleQuota)
-            {
-                throw new AzureBatchQuotaMaxedOutException($"No remaining active jobs quota available. There are {batchUtilization.ActivePoolsCount} active jobs out of {batchVmFamilyBatchQuotas.ActiveJobAndJobScheduleQuota}.");
-            }
+            var (_, exception) = CheckBatchAccountPoolOrJobQuotasImpl(required: 1, batchUtilization.ActiveJobsCount, batchUtilization.ActivePoolsCount, batchVmFamilyBatchQuotas.ActiveJobAndJobScheduleQuota, batchVmFamilyBatchQuotas.PoolQuota);
 
-            if (batchUtilization.ActivePoolsCount + 1 > batchVmFamilyBatchQuotas.PoolQuota)
+            if (exception is not null)
```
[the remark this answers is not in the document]
BMurri: The caller will throw the returned exception for every task assigned to every doomed new pool request (so could be lots). It's much more efficient to return the exception to selectively throw it for a subset of tasks rather than catch it to selectively rethrow it for that same subset of tasks.

The NET standard libraries do this "all the time"

BMurri:
@giventocode This will be ready for a re-review by the time you're expected back in the office.

src/TesApi.Web/Events/RunnerEventsProcessor.cs
```diff
@@ -0,0 +1,516 @@
...
+        /// <remarks>This method assumes every <see cref="RunnerEventsMessage"/> was successfully validated by <see cref="ValidateMessageMetadata(RunnerEventsMessage)"/>.</remarks>
+        public IEnumerable<T> OrderProcessedByExecutorSequence<T>(IEnumerable<T> source, Func<T, RunnerEventsMessage> messageGetter)
+        {
+            ArgumentNullException.ThrowIfNull(source);
+            ArgumentNullException.ThrowIfNull(messageGetter);
+
+            return source.OrderBy(t => OrderBy(messageGetter(t))).ThenBy(t => ThenBy(messageGetter(t)));
+
+            static DateTime OrderBy(RunnerEventsMessage message)
+                => (message.RunnerEventMessage?.Created ?? DateTime.Parse(message.Tags["created"])).ToUniversalTime();
```
giventocode, line 237: The created property should never be null, so we should declare the type as such.

src/TesApi.Web/Events/RunnerEventsProcessor.cs
```diff
@@ -0,0 +1,516 @@
...
+            await azureProxy.SetBlobTags(
+                message.BlobUri,
+                message.Tags
+                    .Append(new(ProcessedTag, DateTime.UtcNow.ToString("O")))
+                    .ToDictionary(pair => pair.Key, pair => pair.Value),
+                cancellationToken);
+        }
+
+        /// <summary>
+        /// Prevents the message from being reprocessed by removing the 'task-id' tag. Used for malformed blobs.
```
giventocode, line 480: What is considered a malformed blob?

src/TesApi.Web/AzureProxy.cs
```diff
@@ -499,59 +292,71 @@ public async Task<string> GetStorageAccountKeyAsync(StorageAccountInfo storageAc
...
         /// <inheritdoc/>
         public Task UploadBlobAsync(Uri blobAbsoluteUri, string content, CancellationToken cancellationToken)
-            => new CloudBlockBlob(blobAbsoluteUri).UploadTextAsync(content, null, null, null, null, cancellationToken);
+            => new BlobClient(blobAbsoluteUri, new(BlobClientOptions.ServiceVersion.V2021_04_10))
+                .UploadAsync(BinaryData.FromString(content), options: null, cancellationToken);
 
         /// <inheritdoc/>
         public Task UploadBlobFromFileAsync(Uri blobAbsoluteUri, string filePath, CancellationToken cancellationToken)
-            => new CloudBlockBlob(blobAbsoluteUri).UploadFromFileAsync(filePath, null, null, null, cancellationToken);
+            => new BlobClient(blobAbsoluteUri, new(BlobClientOptions.ServiceVersion.V2021_04_10))
```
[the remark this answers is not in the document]
BMurri: Yes

src/TesApi.Web/AzureProxy.cs
```diff
@@ -499,59 +292,71 @@ public async Task<string> GetStorageAccountKeyAsync(StorageAccountInfo storageAc
...
         /// <inheritdoc/>
-        public async Task<IEnumerable<CloudBlob>> ListBlobsAsync(Uri directoryUri, CancellationToken cancellationToken)
+        public IAsyncEnumerable<BlobNameAndUri> ListBlobsAsync(Uri directoryUri, CancellationToken cancellationToken)
         {
-            var blob = new CloudBlockBlob(directoryUri);
-            var directory = blob.Container.GetDirectoryReference(blob.Name);
+            var directory = new BlobClient(directoryUri, new(BlobClientOptions.ServiceVersion.V2021_04_10));
+            return directory.GetParentBlobContainerClient()
+                .GetBlobsAsync(prefix: directory.Name.TrimEnd('/') + "/", cancellationToken: cancellationToken)
+                .Select(blobItem => new BlobNameAndUri(blobItem.Name, new BlobUriBuilder(directory.Uri) { Sas = null, BlobName = blobItem.Name }.ToUri()));
```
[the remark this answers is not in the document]
BMurri: Yes

src/TesApi.Web/BatchPool.cs
```diff
@@ -483,38 +461,41 @@ internal enum ServiceKind
...
-                    case ComputeNodeState.Running:
-                    case ComputeNodeState.Creating:
-                    case ComputeNodeState.Starting:
-                    case ComputeNodeState.WaitingForStartTask:
-                        return false;
-                }
+                return false;
             }
 
+            //await foreach (var node in _azureProxy.ListComputeNodesAsync(PoolId, new ODATADetailLevel(selectClause: "state")).WithCancellation(cancellationToken))
```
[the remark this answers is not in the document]
BMurri: It's been a while.

No, we don't need it. I'll remove it.

src/TesApi.Web/Events/RunnerEventsProcessor.cs
BMurri: `ArgumentException` (or derived) thrown from either `ValidateMessageMetadata`, `DownloadAndValidateMessageContentAsync`, or `GetMessageBatchStateAsync` (arguably only ValidateMessageMetadata() counts towards that definition.

Looking at the implementation, I think it would be better to either add the "processed" tag (or replace the last tag with one if there isn't room to add one) with a value like "error" or an empty string. I'll change that.

src/TesApi.Web/Events/RunnerEventsProcessor.cs
```diff
@@ -0,0 +1,516 @@
...
+            }
+            catch (Exception ex)
+            {
+                throw new DownloadOrParseException($"Event message blob is malformed. {ex.GetType().FullName}:{ex.Message}", ex);
+            }
+
+            message = new(message, content);
+
+            // Validate content
+            Validate(Guid.TryParse(content.Id, out _),
```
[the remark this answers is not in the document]
BMurri: Because I was stress testing, I was looking for every possible place I could conceive things could break. Certainly, these two sections can be relaxed.

src/Tes.Runner/Transfer/BlobUploader.cs
```diff
@@ -155,6 +166,11 @@ public virtual async Task<long> UploadAsync(List<UploadInfo> uploadList)
         {
             ValidateUploadList(uploadList);
 
+            foreach (var upload in uploadList.Where(upload => !string.IsNullOrWhiteSpace(upload.MountParentDirectory)).Where(upload => upload.FullFilePath.StartsWith(upload.MountParentDirectory!)))
```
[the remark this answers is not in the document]
BMurri: TES spec v1.0: [tesOutputFileLog](https://github.com/ga4gh/task-execution-schemas/blob/9e9c5aa2648d683d5574f9dbd63a025b4aea285d/openapi/task_execution_service.openapi.yaml#L466)

src/Tes.Runner/Transfer/BlobUploader.cs
```diff
@@ -14,6 +14,9 @@ public class BlobUploader : BlobOperationPipeline
     {
         private readonly ConcurrentDictionary<string, Md5HashListProvider> hashListProviders = new();
 
+        internal readonly IDictionary<string, int> mapPathToMountPrefixLength = new Dictionary<string, int>();
```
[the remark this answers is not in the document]
BMurri: This is written only in one place at only one time from only one thread, and it is then read later on after `UploadAsync` has completely completed in only one other place by only one thread.

src/Tes.Runner/Transfer/BlobUploader.cs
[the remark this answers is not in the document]
BMurri: This method reaches this point only one time in each instance in just one thread, at which point no other thread is accessing either dictionary.

src/Tes.Runner/Executor.cs
```diff
@@ -126,7 +128,7 @@ public async Task<long> UploadOutputsAsync(BlobPipelineOptions blobPipelineOptio
 
                 var optimizedOptions = OptimizeBlobPipelineOptionsForUpload(blobPipelineOptions, outputs);
 
-                bytesTransferred = await UploadOutputsAsync(optimizedOptions, outputs);
+                (bytesTransferred, completedFiles) = await UploadOutputsAsync(optimizedOptions, outputs);
```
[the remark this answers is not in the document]
BMurri: I will

src/Tes.Runner/Executor.cs
```diff
@@ -137,23 +139,24 @@ public async Task<long> UploadOutputsAsync(BlobPipelineOptions blobPipelineOptio
...
             }
             finally
             {
-                await eventsPublisher.PublishUploadEndEventAsync(tesNodeTask, numberOfOutputs, bytesTransferred, statusMessage, errorMessage);
+                await eventsPublisher.PublishUploadEndEventAsync(tesNodeTask, numberOfOutputs, bytesTransferred, statusMessage, errorMessage, completedFiles);
             }
         }
 
-        private async Task<long> UploadOutputsAsync(BlobPipelineOptions blobPipelineOptions, List<UploadInfo> outputs)
+        private async Task<(long BytesTransferred, IEnumerable<CompletedUploadFile> CompletedFiles)> UploadOutputsAsync(BlobPipelineOptions blobPipelineOptions, List<UploadInfo> outputs)
```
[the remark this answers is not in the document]
BMurri: I will

src/TesApi.Web/Events/RunnerEventsProcessor.cs
BMurri: Defense in depth. Doesn't cost much to never run code that won't be hit.

src/Tes.Runner/Executor.cs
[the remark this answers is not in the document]
BMurri: Done

src/Tes.Runner/Executor.cs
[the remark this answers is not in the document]
BMurri: Done

src/TesApi.Web/BatchPool.cs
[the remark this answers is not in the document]
BMurri: Done

Issue #437, opened 2023-10-01
[#148 Implement TES Task Runner] Event publishing by the TES runner


giventocode:
**Problem.**
The TES server cannot access the status of Task processing except through the Batch APIs. This causes delays in completing tasks. The Batch APIs often mark Job/Tasks as completed much later than their actual completion time. Moreover, querying the Batch API for state leads to throttling issues and more delays when scaling up

**Solution:**
The TES runner controls the execution of all the steps in the Batch node. It can generate events that describe the execution and send them to a sink/destination for consumption.

**AC:**
  -  Events must be published to a storage account 
    - This will mitigate the need for additional infrastructure to run TES on Azure and Terra.
    - Must faciliate manual and programmatic discovery/consumption of the messages.
      -  Must use the following path-pattern to publish the events:
          -  /tes-internal/events/{event-name}/YYYY/MM/DD/{timestamp}{workflow_id}{task_id}.json
      - Must set tags in each blob for: {task_id}, {workflow_id}, {event_name} and {created}.
         - Created tag must use HTTP GET friendly format, e.g. [ISO 8601](http://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations) 
  - Must create events for the following actions: download start/end, execution start/end, upload start/end and completion.
    - Download and upload events must include number of files and total bytes transferred. 
    - Completion event must indicate success or failure. 
  - Events must follow the same schema,  and enable correlation to the tasks and workflows.

ngambani:
@giventocode can we link this issue to the TES runner https://github.com/microsoft/ga4gh-tes/issues/148? Or should this be separate from the runner implementation?

Issue #438, opened 2023-10-02
Improve node execution stdout and stderr publishing 


giventocode:
**Problem:**
Stdout/stderr files are only accessible after the task is finished. This makes it hard to troubleshoot long running tasks. Batch uploads these files using SAS tokens that the TES server generates beforehand. In Terra, these tokens might expire if the workflows take too long to execute

**Solution:**
Publish stdout and stderr data to Azure storage as they are being produced by the runner.  

**AC:**
 - Create separate files for each operation: download-stdout.txt, execution-stdout.txt, upload-stdout.txt, download-stderr.txt, execution-stderr.txt, and upload-stderr.txt.
  - Use append blobs and be aware of their limits (4 MiB per block and 50,000 blocks maximum).
  - Buffer the data and upload the buffer when it reaches 4 MiB or 30 seconds have passed.
  - Upload operations must be asynchronous and not blocking. 
  - The TES server must set the location of the logs and the transformation strategy to obtain the JIT SAS token.

PR #443, opened 2023-10-04
[437] Runner event publishing and storage sink

giventocode:
Closes: #437 

In this PR:
- The runner generates events according to #437.
  - The event generation is non-blocking (runs asynchronously in the background), reliable (attempts to retry with reasonable timeouts), and silent (does not affect processing if it fails).
- The TES server sets where events must be sent depending on runtime environment - Terra or CoA.
  -  Destination is set by a new property in the node task definition,  `RuntimeOptoins.StorageEventSink.TargetUrl` with its transformation strategy.
- The deployer now sets Data Owner role to the MI - Data Owner is required for setting tags, when using a delegated key to generate the SAS token. 
- Additional fixes/changes:
  - Removed Docker clean up options from the node task definition model as it is no longer used.   
  - Refactoring of `DockerExecutor` and `MultiplexedStreamLogReader` to tests scenarios within the scope of this issue, and facilitate the implementation of #438


src/Tes.Runner/Events/BlobStorageEventSink.cs
```diff
@@ -0,0 +1,79 @@
...
+using Tes.Runner.Transfer;
+using static System.DateTime;
+
+namespace Tes.Runner.Events
+{
+    public class BlobStorageEventSink : EventSink
+    {
+        const string EventTimeStampFormat = "HH-mm-ss.fff";
+        const string Iso8601DateFormat = "yyyy-MM-ddTHH:mm:ss.fffZ";
+        private const string ApiVersion = "2023-05-03";
```
MattMcL4475, line 16: would be helpful to add a comment with a link to where you looked this up

src/Tes.Runner/Events/BlobStorageEventSink.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        private string ToEventUrl(Uri uri, EventMessage message)
+        {
+            var blobBuilder = new BlobUriBuilder(uri);
+
+            var blobName = ToBlobName(message);
+
+            if (!string.IsNullOrWhiteSpace(blobBuilder.BlobName))
+            {
+                blobName = $"{blobBuilder.BlobName.TrimEnd('/')}/{blobName}";
```
MattMcL4475, line 53: blobName.TrimStart('/') to be safe too?

src/Tes.Runner/Events/EventSink.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            return new Dictionary<string, string>
+            {
+                { "event_name", eventMessage.Name },
+                { "event_id", eventMessage.Id },
+                { "entity_type", eventMessage.EntityType },
+                { "task_id", eventMessage.EntityId },
+                { "workflow_id", eventMessage.CorrelationId },
+                //format date to ISO 8601, which is URL friendly
+                { "created", eventMessage.Created.ToString("yyyy-MM-ddTHH:mm:ssZ") }
```
MattMcL4475, line 56: didn't you already create a const for this?

`const string Iso8601DateFormat`

src/Tes.Runner/Events/EventSink.cs
```diff
@@ -0,0 +1,82 @@
...
+            };
+        }
+        private async Task EventHandlerAsync()
+        {
+            while (await events.Reader.WaitToReadAsync())
```
MattMcL4475, line 61: could this ever throw?  would it be better to put all of this inside a while(true){try{}catch{} await Task.Delay(1000);} equivalent so an exception will never crash the method?

src/Tes.Runner/Events/EventMessage.cs
```diff
@@ -0,0 +1,40 @@
...
+    [JsonPropertyName("created")]
+    public DateTime Created { get; set; }
+    [JsonPropertyName("eventVersion")]
+    public string EventVersion { get; set; } = null!;
+
+    [JsonPropertyName("eventDataVersion")]
+    public string EventDataVersion { get; set; } = null!;
+
+    [JsonPropertyName("eventData")]
+    public Dictionary<string, string>? EventData { get; set; }
```
MattMcL4475, line 39: = new Dictionary<string, string>(); // see comment in EventSink.cs

src/Tes.Runner/Events/EventsPublisher.cs
```diff
@@ -0,0 +1,225 @@
...
+using Tes.Runner.Storage;
+using Tes.Runner.Transfer;
+
+namespace Tes.Runner.Events;
+
+public class EventsPublisher : IAsyncDisposable
+{
+    const string EventVersion = "1.0";
+    const string EventDataVersion = "1.0";
+    public const string TesTaskRunnerEntityType = "TesRunnerTask";
```
MattMcL4475, line 16: Should we add all of these to a Wiki page docs?

src/Tes.Runner/Events/EventsPublisher.cs
```diff
@@ -0,0 +1,225 @@
...
+    public const string DownloadStartEvent = "downloadStart";
```
MattMcL4475, line 17: Should we add all of these to a Wiki page docs?

src/Tes.Runner/Events/EventsPublisher.cs
```diff
@@ -0,0 +1,225 @@
...
+            EventDataVersion = EventDataVersion
+        };
+    }
+
+
+    private async Task PublishAsync(EventMessage message)
+    {
+        if (sinks.Count == 0)
+        {
+            logger.LogInformation("No sinks configured for publishing events");
```
MattMcL4475, line 201: LogWarning?

src/Tes.Runner/Events/EventSink.cs
```diff
@@ -0,0 +1,82 @@
...
+
+using System.Threading.Channels;
+using Microsoft.Extensions.Logging;
+using Tes.Runner.Transfer;
+
+namespace Tes.Runner.Events
+{
+    public abstract class EventSink : IEventSink
+    {
+        const int StopWaitDurationInSeconds = 60;
```
MattMcL4475, line 12: I think TimeSpan.FromSeconds(60); would be better here

src/Tes.Runner/Events/EventsPublisher.cs
```diff
@@ -0,0 +1,225 @@
...
+
+            await sink.PublishEventAsync(message);
+        }
+    }
+
+    public async Task FlushPublishersAsync(int waitTimeInSeconds = 60)
+    {
+        var stopTasks = sinks.Select(s => s.StopAsync());
+
+        await Task.WhenAll(stopTasks).WaitAsync(TimeSpan.FromSeconds(waitTimeInSeconds));
```
MattMcL4475, line 217: I think TimeSpan would be better for waitTime, that way never confusion between "FromSeconds" and int

src/Tes.Runner/Events/EventsPublisher.cs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    public async ValueTask DisposeAsync()
+    {
+        GC.SuppressFinalize(this);
```
MattMcL4475, line 222: I don't think this is necessary if there is no ~Destructor

src/Tes.Runner/Events/EventsPublisher.cs
```diff
@@ -0,0 +1,225 @@
...
+        await FlushPublishersAsync();
```
MattMcL4475, line 223: this should be idempotent fyi; not sure if it currently is

src/Tes.Runner/Executor.cs
```diff
@@ -4,42 +4,80 @@
...
 using Tes.Runner.Storage;
 using Tes.Runner.Transfer;
 
 namespace Tes.Runner
 {
-    public class Executor
+    public class Executor : IAsyncDisposable
     {
         public const long ZeroBytesTransferred = 0;
+        public const long DefaultErrorExitCode = -1;
```
MattMcL4475, line 17: In Linux, exit codes can only be 0-255, so it would be better to use "1" as the error exit code here

src/Tes.Runner/Transfer/BlobBlockApiHttpUtils.cs
```diff
@@ -5,6 +5,7 @@
 using System.Net.Http.Headers;
 using System.Net.Sockets;
 using System.Text;
+using System.Web;
```
MattMcL4475, line 8: I don't think we want this dependency in ASP.NET Core... do we really need System.Web ?

src/Tes.Runner/Transfer/BlobBlockApiHttpUtils.cs
```diff
@@ -45,6 +46,31 @@ public static HttpRequestMessage CreatePutBlockRequestAsync(PipelineBuffer buffe
...
+        return request;
+    }
+
+    private static void AddPutBlobHeaders(HttpRequestMessage request, string apiVersion, Dictionary<string, string>? tags)
+    {
+        AddPutBlockHeaders(request, apiVersion);
+
+        if (tags is { Count: > 0 })
+        {
+            var tagsHeader = string.Join("&", tags.Select(t => $"{t.Key}={HttpUtility.UrlEncode(t.Value)}"));
```
MattMcL4475, line 68: you can use System.Net.WebUtility.UrlEncode instead so we don't need System.Web

src/Tes.Runner/Events/EventSink.cs
giventocode: I don't think it is necessary. I can't think of a realistic scenario where checking if the channel has data would throw. Also, once the channel reaches an end/error state, keep trying in a forced loop would not help. However, if you can think of a scenario this method can throw, let's discuss it and see what is the best way to mitigate it.

src/Tes.Runner/Events/EventSink.cs
```diff
@@ -0,0 +1,82 @@
...
+                        logger.LogDebug($"Handling event. Event Name: {eventMessage.Name} Event ID: {eventMessage.Id} ");
+
+                        await HandleEventAsync(eventMessage);
+
+                        logger.LogDebug($"Event handled. Event Name: {eventMessage.Name} Event ID: {eventMessage.Id} ");
+                    }
+                    catch (Exception e)
+                    {
+                        //event error should be handled silently, as event failures should not affect overall processing....
+                        logger.LogError(e, $"Error handling event. Event ID: {eventMessage.Id}");
```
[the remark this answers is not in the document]
giventocode: I think the risk of creating logs in a loop is low. Each task should generate at most 7 events, so the logging amount will be small. I also dislike eating the exception. I wonder if we should stop the processing completely when we encounter an exception at this point.

src/Tes.Runner/Events/EventsPublisher.cs
giventocode: Yes, I agree.

MattMcL4475 (approved):

src/Tes.Runner/Events/BlobStorageEventSink.cs
giventocode: No need, `ToBlobName` always returns a value that starts with `events/...`

src/Tes.Runner/Events/EventSink.cs
MattMcL4475: Great just wanted to check.  Maybe any of these?

`ObjectDisposedException`: If the underlying channel is disposed of before or during the operation.
`ChannelCompletionException`: If the channel has been marked as complete with an exception, then waiting to read could propagate that exception.
`OperationCanceledException`: If the CancellationToken passed to it gets canceled.
Other unexpected exceptions: In the case of bugs, resource issues, or unexpected behavior, other exceptions might be thrown, although these are not typical for this operation.

src/Tes.Runner/Events/EventsPublisher.cs
giventocode: No sure I see the benefit here, also I don't think it is possible to create a constant of type TimeSpan set to 60s.

src/Tes.Runner/Events/EventsPublisher.cs
giventocode: yes, that is correct. Good catch, thank you - added by the refactoring tool I used.

src/Tes.Runner/Transfer/BlobBlockApiHttpUtils.cs
giventocode: URL encoding for the tags.

src/Tes.Runner/Events/EventsPublisher.cs
giventocode: Yep, it is, but I'll double check it.

src/Tes.Runner/Events/EventMessage.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation.
```
BMurri, line 1: It would be extremely useful if this file were in a project that both `Tes.Runner` and `TesApi.Web` both share, along with an enumeration of the various events currently being const strings in `EventsPublisher.cs`

BMurri (approved):

src/Tes.Runner/Events/EventsPublisher.cs
MattMcL4475: np, not a big deal, "TimeSpan is often preferred for readability and type safety when working with time-related operations. It makes the code self-documenting, and there's less chance of unit errors (e.g., mistakenly treating the delay as milliseconds rather than seconds)."

MattMcL4475: i.e. `private static TimeSpan stopWaitDuration = TimeSpan.FromSeconds(60);`

PR #455, opened 2023-10-11
Use Channel to reduce write times

### LINKED ISSUE #456: Database writes could be reduced below 50ms
Currently database writes take about 50ms, which is plenty fast but could be optimized even further with thread synchronization.

MattMcL4475:
Uses `async`/`await` semantics with a `Channel` and eliminates the need for a polling loop that uses `Task.Delay`

Resolves #456

### DIFF
--- a/src/Tes/Repository/PostgreSqlCachingRepository.cs
+++ b/src/Tes/Repository/PostgreSqlCachingRepository.cs
@@ -117,7 +117,7 @@ namespace Tes.Repository
         /// <param name="item"></param>
         /// <param name="action"></param>
         /// <returns></returns>
-        protected async Task<T> AddUpdateOrRemoveItemInDbAsync(T item, WriteAction action, CancellationToken cancellationToken)
+        protected Task<T> AddUpdateOrRemoveItemInDbAsync(T item, WriteAction action, CancellationToken cancellationToken)
         {
             var source = new TaskCompletionSource<T>();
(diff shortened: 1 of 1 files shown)


MattMcL4475:
Reuslts in a 5.86x speedup, for an average write time of 8.19ms (vs. 48ms)
image

src/Tes/Repository/PostgreSqlCachingRepository.cs
```diff
@@ -134,8 +134,8 @@ protected Task<T> AddUpdateOrRemoveItemInDbAsync(T item, WriteAction action, Can
                 }
             }
 
-            _itemsToWrite.Enqueue((item, action, source));
-            return result;
+            await _itemsToWrite.Writer.WriteAsync((item, action, source));
+            return item;
```
MattMcL4475, line 138: Do we need to add `await result;` above this line?

BMurri: I would revert lines 120 and 138 and remove the await on 137. The purpose of the implementation of this method is to return a Task<T> representing the writing of the task to the DB.

If you insist on having this method be async (overall the whole system will be faster and have smaller code if it is not) then you should replace line 138 with `var returnValue = await result;\n return returnValue;`

BMurri (approved):
Looks good overall. Please consider my comment.

PR #460, opened 2023-10-13
Add an option to specify a container for SupportedBackendParameters.internal_path_prefix paths

jsaun:
Cromwell currently returns different paths depending if the blob filesystem is used or the local filesystem. See: https://github.com/broadinstitute/cromwell/pull/7208/files . This was done to fix an issue in Tes where DefaultStorageAccessProvider and TerraStorageAccessProvider are actually constructing the paths differently. 

This PR will accommodate the container name not being included in the TesResources.SupportedBackendParameters.internal_path_prefix property. In CoA, we will have to set Storage__ExecutionsContainerName to cromwell-executions to use the blobfilesystem. 

Part of https://github.com/microsoft/CromwellOnAzure/issues/637

### DIFF
--- a/src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
+++ b/src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
@@ -205,8 +205,16 @@ namespace TesApi.Web.Storage
             if (task.Resources?.ContainsBackendParameterValue(TesResources.SupportedBackendParameters
                     .internal_path_prefix) == true)
             {
+                var internalPath = $"{task.Resources.GetBackendParameterValue(TesResources.SupportedBackendParameters.internal_path_prefix).Trim('/')}{normalizedBlobPath}";
+                
+                if (storageOptions.ExecutionsContainerName is not null &&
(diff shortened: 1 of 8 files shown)


src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
```diff
@@ -204,8 +205,16 @@ public override async Task<string> GetInternalTesTaskBlobUrlAsync(TesTask task,
             if (task.Resources?.ContainsBackendParameterValue(TesResources.SupportedBackendParameters
                     .internal_path_prefix) == true)
             {
+                var internalPath = $"{task.Resources.GetBackendParameterValue(TesResources.SupportedBackendParameters.internal_path_prefix).Trim('/')}{normalizedBlobPath}";
+
+                if (storageOptions.ExecutionsContainerName is not null &&
+                    !internalPath.StartsWith(storageOptions.ExecutionsContainerName, StringComparison.OrdinalIgnoreCase))
```
BMurri, line 211: Would this be better?
internalPath.StartsWith(storageOptions.ExecutionsContainerName + "/",

PR #483, opened 2023-10-24
Address autoscale formula evaluation errors

### LINKED ISSUE #482: Batch should not be returning InsufficientSampleData errors
**Describe the bug**
Our autoscale formulas are written to never return evaluation errors if Batch is running according to the docs. Experience shows that errors are returned, and when they are, the pools get "stuck".

**Steps to Reproduce**
Run workflows. It'll happen sooner or later.

**Expected behavior**
TES should attempt to reset the scaling on the pool in an effort to "unstick" it.

**Additional context**
Start with the minimal code changes to effect this fix.
Expand later if this exposes other issues.

BMurri:
fixes #482

In brief: The current code resets pools with long breaks in the reevaluation of their auto-scale formulas. This PR does 2 things: adds an additional test to also rest pools with auto-scale formula error results, and fixes and clarifies the flow of the state machine managing the health of the pools.

Changes:
- Added comments describing the state machine and places where Batch does something unexpected.
- Added new test for this PR's issue
- Changed ServicePoolGetResizeErrorsAsync() to collect pool error states no matter the state of the state machine.
- Changed ServicePoolGetResizeErrorsAsync() to separate collection of autoscale formula evaluation errors (which don't require pool resizes to occur) from resize error collections.
- Changed ServicePoolGetResizeErrorsAsync() to use the AllocationStateTransitionTime rather than the expected longevity of a pool resize to gate resize error collections.

Still TODO (considering for a separate PR, as these tasks will need to be done through `BatchScheduler`):
- Transfer queued `CloudTask`s assigned to unresponsive pools to a new pool with the same key.
- Delete unresponsive pools when their running task count drops to zero.


src/TesApi.Web/BatchPool.cs
```diff
@@ -72,107 +71,190 @@ private async ValueTask RemoveNodesAsync(IList<ComputeNode> nodesToRemove, Cance
...
 
         private DateTime? Creation { get; set; }
+        private DateTime? AllocationStateTransitionTime { get; set; }
         private bool IsDedicated { get; set; }
 
         private void EnsureScalingModeSet(bool? autoScaleEnabled)
         {
-            if (ScalingMode.Unknown == _scalingMode)
+            /*
+             * If the the scaling mode does not correspond to the actual state of autoscale enablement, this method guides us towards nivana.
```
MattMcL4475, line 143: "the desired state" would be more appropriate

MattMcL4475 (approved):

Issue #485, opened 2023-10-26
Runner fails to download/upload blobs when running on Terra


giventocode:
**Describe the bug**
The runner cannot access the storage account to download or upload data, or to send logs, when it runs on Terra.

**Steps to Reproduce**
Run a workflow with inputs and outputs on Terra

**Additional context**
Cromwell passes to TES the identity name rather than a resource id in the workflow identity property of the Task. If Cromwell used the resource id, it would cause an error in the WSM call to create the batch pool. The WSM API requires the identity name.

PR #486, opened 2023-10-27
[485] Added support for workflow_identity to provide a Managed Identity (MI) resource ID or MI name when running on Terra

giventocode:
Fixes: #485 

In this PR:

- If the workflow identity has a value that is not a valid Azure resource ID, `TaskToNodeTaskConverter` creates a resource ID for the runner's managed identity (MI) using the BatchAccount's subscription and resource group.
  - The `TaskToNodeTaskConverter` assumes the value in the workflow identity is the name of the MI resource.  
 
- If the workflow identity has a value that is a valid Azure resource ID for the MI, `TaskToNodeTaskConverter` assigns that value to the runner's MI resource ID.
  - When calling WSM to create a batch pool, the identity name is extracted from the resource ID. 

- If the workflow identity is not specified, the `TaskToNodeTaskConverter` uses the global MI setting as a fallback.

- Added validation to the `NodeTaskBuilder` to ensure that only valid Azure resource IDs are allowed in the runner task.

### DIFF
--- a/src/TesApi.Web/Runner/TaskToNodeTaskConverter.cs
+++ b/src/TesApi.Web/Runner/TaskToNodeTaskConverter.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Azure.Storage.Blobs;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Tes.Extensions;
 using Tes.Models;
 using Tes.Runner.Models;
 using TesApi.Web.Management.Configuration;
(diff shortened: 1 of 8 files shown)


src/TesApi.Web/Runner/TaskToNodeTaskConverter.cs
```diff
@@ -400,7 +433,6 @@ private static string AppendParentDirectoryIfSet(string inputPath, string pathPa
     /// </summary>
     /// <param name="AdditionalInputs"></param>
     /// <param name="DefaultStorageAccountName"></param>
-    /// <param name="NodeManagedIdentityResourceId"></param>
```
BMurri, line 403: Pretty sure this line should be `/// <param name="GlobalManagedIdentity"></param>` rather than removed.

BMurri (approved):

Issue #489, opened 2023-10-27
Number of files is zero in the downloadEnd event.


giventocode:
**Describe the bug**
The number of files property is always zero in the downloadEndEvent.
E.g. : `"eventData":{"numberOfFiles":"0","totalSizeInBytes":"2195","errorMessage":""}`

**Steps to Reproduce**
A TES tasks with inputs to download

**Expected behavior**
`numberOfFiles` should match the number of inputs.

PR #490, opened 2023-10-27
[489] Executor sets the number of files for event data

giventocode:
Fixes: #489 

In this PR:
 - The executor sets the number of files for the event data correctly.
 - Added factory abstraction for the creation of uploader and downloaders inside executor, to faciliate testing. 
 - The test validates the correction.

### DIFF
--- a/src/Tes.Runner.Test/ExecutorTests.cs
+++ b/src/Tes.Runner.Test/ExecutorTests.cs
@@ -19,6 +19,9 @@ namespace Tes.Runner.Test
         private BlobPipelineOptions blobPipelineOptions = null!;
         private Mock<EventsPublisher> eventsPublisherMock = null!;
         private Mock<DockerExecutor> dockerExecutorMock = null!;
+        private Mock<ITransferOperationFactory> transferOperationFactoryMock = null!;
+        private Mock<BlobDownloader> blobDownloaderMock = null!;
+        private Mock<BlobUploader> blobUploaderMock = null!;
 
         [TestInitialize]
         public void SetUp()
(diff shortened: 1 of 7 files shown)


BMurri (approved):

src/Tes.Runner.Test/ExecutorTests.cs
```diff
@@ -79,6 +92,33 @@ public async Task DownloadInputsAsync_InputProvided_StartSuccessEventsAreCreated
...
+            };
+
+            blobDownloaderMock.Setup(r => r.DownloadAsync(It.IsAny<List<DownloadInfo>>()))
+                .ReturnsAsync(0);
+
+            fileOperationResolverMock.Setup(r => r.ResolveInputsAsync()).ReturnsAsync(inputs);
+            var result = await executor.DownloadInputsAsync(blobPipelineOptions);
+            Assert.AreEqual(Executor.ZeroBytesTransferred, result);
+            eventsPublisherMock.Verify(p => p.PublishDownloadStartEventAsync(It.IsAny<NodeTask>()), Times.Once);
+            eventsPublisherMock.Verify(p => p.PublishDownloadEndEventAsync(It.IsAny<NodeTask>(), nodeTask.Inputs.Count, 0, EventsPublisher.SuccessStatus, string.Empty), Times.Once);
```
MattMcL4475, line 119: Maybe add another test or modify this one, to test a non-zero number of inputs and assert the download count, since that was the original bug?

MattMcL4475 (approved):

src/Tes.Runner.Test/ExecutorTests.cs
giventocode: Hmm this is what this test is asserting. It verifies that the event is created with the correct number of files (nodeTask.Inputs.Count).

PR #503, opened 2023-11-10
Log retries

### LINKED ISSUE #502: Log retries
**Problem:**
Sometimes, a problem cannot be diagnosed without access to the information returned by APIs that were retried.

**Solution:**
Logging is the most convenient place to store information received during retries. Making it easy to hook into the Polly workflow would facilitate that effort.

Trolling API error reports. Not very informative WRT the reasons for many of the failures.

**Code dependencies**
* `Build pipeline?` No
* `Integration tests?` No

**Additional context**
PR will accompany this issue.

BMurri:
fixes #502

Given changes made to the deployers, I made the relevant classes available to them as well (although they are only using one part of code that moved that is not tied to retry logging).

The only part of the node runner that are now logging retries with this solution that I'm aware are the runner's WSM calls and docker image pulls.

I've somewhat minimized variable name changes (even though other names would be better) to better illustrate the impact of this solution to the problem. TBD if I rename those variables before or after submitting this to main.

I've also refactored `HttpApiClient` a little bit to have simpler multi-purpose flows and make it easier to know which method to use depending on your needs (it had gotten a little chaotic before, which is normal in pretty much every project) as well as minimize the amount of duplicated code all over the place.


src/Tes.ApiClients/CachingRetryHandler.cs
```diff
@@ -43,12 +44,13 @@ public CachingRetryHandler(IMemoryCache appCache, IOptions<RetryPolicyOptions> r
         /// <param name="cacheKey"></param>
         /// <param name="action">Action to execute</param>
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> for controlling the lifetime of the asynchronous operation.</param>
+        /// <param name="context"></param>
         /// <returns></returns>
-        public virtual async Task<TResult> ExecuteWithRetryAndCachingAsync<TResult>(string cacheKey, Func<CancellationToken, Task<TResult>> action, CancellationToken cancellationToken)
+        public virtual async Task<TResult> ExecuteWithRetryAndCachingAsync<TResult>(string cacheKey, Func<CancellationToken, Task<TResult>> action, CancellationToken cancellationToken, Context? context = default)
```
giventocode, line 49: I think exposing the Polly context to the caller is unnecessary, as it does not provide much benefit, and it leads to many changes that we have to maintain for every API call. We can log the retries without passing the context to the caller, as the runner already does with its policy - [see this](https://github.com/microsoft/ga4gh-tes/blob/d470a118a0c448bc55c869c8809324de7468076f/src/Tes.Runner/Transfer/BlobApiHttpUtils.cs#L350). I realize that this runner's retry approach relies on the exception stack to identify the failed method and might produce a verbose log output, unlike the PR approach. However, I prefer this simpler option.

src/Tes.ApiClients/HttpApiClient.cs
```diff
@@ -218,13 +268,11 @@ private async Task<HttpRequestMessage> CreateGetHttpRequest(Uri requestUrl, bool
...
+        protected async Task<string> ReadResponseBodyAsync(HttpResponseMessage response, CancellationToken cancellationToken)
         {
-            var response = await HttpClient.SendAsync(request, cancellationToken);
-
-            response.EnsureSuccessStatusCode();
-
-            return await response.Content.ReadAsStringAsync(cancellationToken);
+            var ctx = new Polly.Context();
+            ctx.SetOnRetryHandler(LogRetryErrorOnRetryHandler());
+            return await cachingRetryHandler.ExecuteWithRetryAsync(response.Content.ReadAsStringAsync, cancellationToken, ctx);
```
giventocode, line 275: Is it valid to retry reading from the stream without making the request again? If the goal here is to retry on socket/io exceptions I believe you must retry by making the request again, not just retrying to read the stream. The runner has a pattern to handle this scenario - by wrapping retriable failures in a retriable exception, which causes the request to be retried.

src/Tes.ApiClients/HttpApiClient.cs
```diff
@@ -68,6 +69,36 @@ protected HttpApiClient(CachingRetryHandler cachingRetryHandler, ILogger logger)
...
+        /// <summary>
+        /// A logging Polly retry handler.
+        /// </summary>
+        /// <typeparam name="TResult">See <see cref="PolicyBuilder{TResult}"/></typeparam>
+        /// <param name="caller">Calling method name.</param>
+        /// <returns><see cref="RetryHandler.OnRetryHandler{TResult}"/></returns>
+        private RetryHandler.OnRetryHandler<TResult> LogRetryErrorOnRetryHandler<TResult>([System.Runtime.CompilerServices.CallerMemberName] string? caller = default)
+            => new((result, timeSpan, retryCount, correlationId) =>
+            {
+                if (result.Exception is null)
```
giventocode, line 92: What are the cases where the exception is null?  I find it kind of weird that Polly would retry without an exception.

src/Tes.ApiClients/CachingRetryHandler.cs
BMurri: I thought of having this be opt-in and flexible, and rather than having to craft unique policies all over the codebase, reusing the ones already provided in `RetryHandler` seemed more maintainable to my mind. It allows for flexibility in extending the policies as well, without restricting usage to only the generally available policies.

What I didn't do was provide a default retry log in-place.

src/Tes.ApiClients/HttpApiClient.cs
BMurri: In my experience it has been very valid for many transient TCP/IP stream issues, as long as the remote connection has not closed the stream (which is a configuration many servers follow to improve performance and reduce memory consumption).

There are only two places in the entire codebase where retrying the entire request due to a failure in reading the stream will not retry the entire request with my changes, and that would be easy to remedy: `Tes.ApiClients.HttpApiClient.HttpGetRequestWithRetryPolicyAsync` and `Tes.ApiClients.HttpApiClient.HttpGetRequestWithCachingAndRetryPolicyAsync`. I'll update the PR so that behavior isn't changed.

src/Tes.ApiClients/HttpApiClient.cs
BMurri: The Polly policy builders that take a generic type argument (like the one returned by `HttpPolicyExtensions.HandleTransientHttpError()`) retriable failures can be defined by state where no exceptions are involved (including conditions added by `OrResult()`). In those cases, exception will be `null` in the retry handler.

src/Tes.ApiClients/HttpApiClient.cs
giventocode: If after your changes a failed stream read will trigger a full request retry, I guess we don't need to retry the ReadAsStringAsync. Am I correct?

BMurri: We don't _need_ to, but if the retry succeeds, we avoid pounding the server with yet another request for basically the same information.

src/Tes.ApiClients/CachingRetryHandler.cs
BMurri: I just took it a step closer: it's still opt-in, but there's no longer any exposing of `Polly.Context`s. It does reduce the line count of the changes overall.

I'm toying with an alternative: making the current `Tes.ApiClients.RetryHandler` into an opinionated retry policy builder based on `RetryOptions` (and extend the builder with constructs from `CachedRetryHandler`). That would pull together all the information needed for the logs without hitting every single line of code where a retry needs to be handled.

src/Tes.ApiClients/HttpApiClient.cs
giventocode: I'd recommend deleting it. This method might not be idempotent or able to "rewind" the underlying the stream if it fails. Even if it works, in some cases, in others (more likely) it could cause a futile retry loop that affects performance.

src/Tes.ApiClients/CachingRetryHandler.cs
giventocode: Sounds good, I agree that moving this functionality to the `CachingRetryHandler` is a better approach.  Also, if making it, opt-in requires many changes, that is something you can defer as in practicality I'd always want to see these retries for troubleshooting.

src/Tes.ApiClients/HttpApiClient.cs
BMurri: Done in my next push to the PR's branch

MattMcL4475 (approved):

PR #514, opened 2023-11-15
Validate requested resources

### LINKED ISSUE #513: Resource requests should be validated when task are created
**Problem:**
If a TesTask is created with a request for zero or a negative number of cpu cores, it's possible that an expensive vm_size might be obtained. Explicit negative or zero quantities of cpu cores, memory, and diskspace are nonsensical and should not be allowed.

**Solution:**
Tasks requesting negative or zero values for disk or memory space or cpu core counts should be rejected as bad requests.

Doing nothing.

**Code dependencies**
* `Build pipeline?` No
* `Integration tests?` No

**Additional context**
Discovered while discussing a finding in the vm sku pricing and availability data.

BMurri:
fixes #513

### DIFF
--- a/src/TesApi.Web/Controllers/TaskServiceApi.cs
+++ b/src/TesApi.Web/Controllers/TaskServiceApi.cs
@@ -187,6 +187,24 @@ namespace TesApi.Controllers
                 tesTask.Resources.Preemptible = cwlDocument.Preemptible ?? tesTask.Resources.Preemptible;
             }
 
+            if (tesTask?.Resources is not null)
+            {
+                if (tesTask.Resources.CpuCores.HasValue && tesTask.Resources.CpuCores.Value <= 0)
+                {
+                    return BadRequest("ram_gb must be greater than zero");
+                }
+
+                if (tesTask.Resources.DiskGb.HasValue && tesTask.Resources.DiskGb.Value <= 0.0)
+                {
+                    return BadRequest("ram_gb must be greater than zero");
+                }
+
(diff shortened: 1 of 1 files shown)


src/TesApi.Web/Controllers/TaskServiceApi.cs
```diff
@@ -187,6 +187,24 @@ public virtual async Task<IActionResult> CreateTaskAsync([FromBody] TesTask tesT
                 tesTask.Resources.Preemptible = cwlDocument.Preemptible ?? tesTask.Resources.Preemptible;
             }
 
+            if (tesTask?.Resources is not null)
+            {
+                if (tesTask.Resources.CpuCores.HasValue && tesTask.Resources.CpuCores.Value <= 0)
+                {
+                    return BadRequest("ram_gb must be greater than zero");
```
MattMcL4475, line 194: `cpu_cores`

src/TesApi.Web/Controllers/TaskServiceApi.cs
```diff
@@ -187,6 +187,24 @@ public virtual async Task<IActionResult> CreateTaskAsync([FromBody] TesTask tesT
...
+                }
+
+                if (tesTask.Resources.DiskGb.HasValue && tesTask.Resources.DiskGb.Value <= 0.0)
+                {
+                    return BadRequest("ram_gb must be greater than zero");
```
MattMcL4475, line 199: `disk_gb`

MattMcL4475 (approved):
Error messages need fixing otherwise LGTM

PR #547, opened 2023-12-13
Remove attempted access to /cromwell-tmp/

### LINKED ISSUE #544: Possible regression
**Describe the bug**
When support was added for CWL files run through Cromwell three or so years ago, TES and Cromwell shared a volume mount (`\cromwell-tmp`) and code was added to TES because Cromwell would supply TES inputs where the `url` was of the form `file:///cromwell-tmp/tmp12345/blob1`.

When we added K8 deployments we did not preserve that volume mapping for TES, and when we subsequently removed support for VM CoA deployments, those inputs would now fail.

Cromwell has now removed support for CWL, but I don't know the history well enough to know if these types of inputs only happen(ed) with CWL, or if Cromwell fixed their code to no longer send inputs pointing to their "tmp" space, or if we're just waiting on latent bug.

BMurri:
Possible resolution to fix #544

### DIFF
--- a/src/TesApi.Tests/BatchSchedulerTests.cs
+++ b/src/TesApi.Tests/BatchSchedulerTests.cs
@@ -1475,41 +1475,6 @@ namespace TesApi.Tests
         }
 
         [TestMethod]
-        [Ignore("Temporary disabled")]
-        public async Task LocalFilesInCromwellTmpDirectoryAreDiscoveredAndUploaded()
-        {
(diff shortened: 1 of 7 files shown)


MattMcL4475 (approved):
LGTM - please run the common workflows integration test to confirm before merging

BMurri:
> LGTM - please run the common workflows integration test to confirm before merging

It passed

PR #562, opened 2023-12-30
Create whitelist when updating VmSize metadata

### LINKED ISSUE #538: VMSize whitelist needs review
**Problem:**
It has been a year since the vmsize whitelist has been updated.

**Solution:**
We should run a new test to see if we can drop the whitelist, and if not, at least update it.

We may consider combining whitelist updates with price list updates, if we can make the whitelist gathering process less painful.

**Code dependencies**
* `Build pipeline?` No
* `Integration tests?` Maybe

**Additional context**
TBA

BMurri:
fixes #538

Notes:
* The old whitelist has been removed. In its place is a validation made by creating pools and allocating one node each for each SKU.
* Failures (and other messages) that are not fatal are now in yellow (fatal errors are in red),
* Configuration is now in a JSON file. The command-line will replace individual entries (and/or add new ones). The previous command-line from the Visual Studio `GenerateBatchVmSkus.csproj.user` is reflected in `appsettings.json`.
* When validation runs, all pools are used simultaneously, and each pool will be used up to the limits of its quotas, in order to minimize the time, it would take to validate each SKU serially. The algorithm isn't perfect, and it assumes that the batch accounts are not running dynamic loads during the validations.
* SKUs are provided to each batch account in the order they appear in the configuration. If there is no condition under which the SKU can be validated (whether because of quota or region) the SKU is sent to the next account for evaluation. Any SKUs left over will be listed at the end.
* Validation pools are given a PoolId that exactly matches the name of the SKU.
* Retriable issues will be retried after 4 minutes to allow the previous attempt's pool to be deleted. After three retries, the SKU will be skipped.
* No tasks are run in the nodes, currently. If the nodes allocated are found either `Idle` or `Preempted`, it is assumed that the SKU will work with CoA/ToA. If the node becomes `Unusable` or if any startup errors are reported, the SKU will be skipped (aka not included).
* Currently no effort is made to balance the work between the different accounts.

Considerations:
* Sometimes the batch-related API will include SKUs that the non-batch-related APIs will not, resulting in a failure. Rerunning usually resolves this.
* Running this tool (with no changes to the code) multiple times in succession will, with some degree of uncertainness, return slightly different results.


src/GenerateBatchVmSkus/appsettings.json
```diff
@@ -0,0 +1,16 @@
+{
+  "SubscriptionId": "594bef42-33f3-42df-b056-e7399d6ae7a0",
+  "OutputFilePath": "BatchSupportedVmSizeInformation.json",
+  "BatchAccounts": [
+    "msfttest52268c",
+    "msftscfc205327",
+    "bmuseast2",
+    "mattmclcoatest"
```
MattMcL4475, line 8: did you mean to check this in?

MattMcL4475 (approved):

src/GenerateBatchVmSkus/appsettings.json
BMurri: `SubscriptionId` and `OutputFilePath` were moved from `src/GenerateBatchVmSkus/Properties/launchSettings.json`. But I could be persuaded that this file should be stored elsewhere and used by whomever runs this tool going forward.

PR #576, opened 2024-01-25
Update to .NET 8 and all Nuget packages

MattMcL4475:
#577


BMurri (approved):
Commit to main as soon as the integration tests pass

PR #582, opened 2024-01-30
Add debug option that enables writing the raw request body to file system for debugging

MattMcL4475:


BMurri (commented):
If this is going to go into main, you should consider a background service to delete them after some amount of time

MattMcL4475:
Thanks!  Just prototyping this now, but was thinking writing to a blob would be even better.

Issue #565, opened 2024-01-10
Add support for AzureUSGovernment, AzureChinaCloud, AzureCloud (default)


MattMcL4475:
- Ability to specify an `AzureCloudName` as a parameter on the command line for the Deployer, such as `AzureCloud`, `AzureUSGovernment`, `AzureChinaCloud`.  
- Ability to specify it in the Helm chart
- Shall enable deploying in any of the respective clouds with functional equivalence

BMurri:
Alternatively, "https://management.azure.com/metadata/endpoints?api-version=2019-03-01" lists all of the clouds with the following names: `AzureCloud`, `AzureChinaCloud`, `AzureUSGovernment`. Pull the host from the `resourceManager` property (it's a URL) and get `https://<resourceManager->Host>/metadata/endpoints?api-version=2023-11-01` and it gives you basically all the URLs listed above, without having to "hardcode" them into our code.

Add a configuration value named `AzureCloudName` to use one of the Azure-run clouds (by name, defaulting to `AzureCloud` or a separate configuration value `AzureCloudUrl` (that is the resource manager URL of the private cloud) to access a private cloud (instead of the first URL above).

BMurri:
Possibly make those `AzureCloud`.`Name` and `AzureCloud`.`Url` to fit with the rest of the Options-enabled configuration.

BMurri:
Examples:

```
{
    "portal": "https://portal.azure.com",
    "authentication": {
        "loginEndpoint": "https://login.microsoftonline.com",
        "audiences": [
            "https://management.core.windows.net/",
            "https://management.azure.com/"
        ],
        "tenant": "common",
        "identityProvider": "AAD"
    },
    "media": "https://rest.media.azure.net",
    "graphAudience": "https://graph.windows.net/",
    "graph": "https://graph.windows.net/",
    "name": "AzureCloud",
    "suffixes": {
        "azureDataLakeStoreFileSystem": "azuredatalakestore.net",
        "acrLoginServer": "azurecr.io",
        "sqlServerHostname": "database.windows.net",
        "azureDataLakeAnalyticsCatalogAndJob": "azuredatalakeanalytics.net",
        "keyVaultDns": "vault.azure.net",
        "storage": "core.windows.net",
        "azureFrontDoorEndpointSuffix": "azurefd.net",
        "storageSyncEndpointSuffix": "afs.azure.net",
        "mhsmDns": "managedhsm.azure.net",
[...]

PR #584, opened 2024-01-31
Add support for the Azure Government cloud

MattMcL4475:
Adds the ability to specify an `AzureCloudName` as a parameter on the command line for the Deployer; accepts: `AzurePublicCloud`, `AzureGovernment`, `AzureChina`.  Can also be set in the Helm chart in AKS.

Resolves: #565


BMurri (commented):
In `src/TesApi.Web/Management/ArmResourceInformationFinder.cs` one of the two methods is called from Program.cs (the other is called from Startup.cs). Those two methods call a third method that creates an azure management client that needs to use the correct cloud. It's missed, and it WILL break logging (if it doesn't prevent the service standing up first).

This is the reason I did the cloud-determination code from Program.cs rather then Startup.cs. But, I have an idea about that wherein I have a better idea.

src/Tes.Runner/Storage/UrlTransformationStrategyFactory.cs
giventocode, line 1: The ARM transformation strategy needs to be updated as well. It determines whether to get the SAS token using MI by looking for the blob URL Azure commercial suffix -which is a constant. It would be simpler to implement logic to get the proper suffix inside the runner if we pass the Azure environment in the runtime options, as suggested earlier. Otherwise, we would have to pass the suffix to look for in the runtime options - which would lead us on the path of config options for each URL needed, which is not ideal.

src/CommonUtilities/Models/NodeTask.cs
```diff
@@ -47,6 +48,8 @@ public class RuntimeOptions
         public StorageTargetLocation? StorageEventSink { get; set; }
 
         public StorageTargetLocation? StreamingLogPublisher { get; set; }
+
+        public AzureCloudIdentityConfig? AzureCloudIdentityConfig { get; set; }
```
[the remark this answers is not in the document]
BMurri: Less work if you just plug-and-play the values here every time.

src/Tes.ApiClients/Tes.ApiClients.csproj
```diff
@@ -28,4 +28,5 @@
     </AssemblyAttribute>
   </ItemGroup>
 
+
```
BMurri, line 31: Remove this blank line, and thus this file, from the PR?

src/Tes.Runner/Storage/UrlTransformationStrategyFactory.cs
BMurri: There's only ever one suffix to look for globally, since different clouds cannot be in the same tenant.

src/TesApi.Web/Options/GeneralOptions.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace TesApi.Web.Options
+{
+    /// <summary>
+    /// General TES system options
+    /// </summary>
+    public class GeneralOptions
```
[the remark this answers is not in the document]
MattMcL4475: I'd rather leave it in General, since no plans for this to ever work on other clouds, and it's such a basic and fundamental setting

src/CommonUtilities/Models/NodeTask.cs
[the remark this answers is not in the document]
MattMcL4475: Maybe change it in the future

src/TesApi.Web/BatchSupportedVmSizeInformation_AZURECHINACLOUD.json
BMurri, line 1: Please open a new issue to rebuild this file using an actual China cloud credential (similar to how the USGov cloud file was built)

src/deploy-tes-docker-image.sh
```diff
@@ -6,20 +6,28 @@
 
 RESOURCE_GROUP_NAME=$1
 ACR_NAME=$2
+IS_US_GOVERNMENT=$3
```
BMurri, line 9: Consider CLOUD_NAME instead of booleans (to facilitate China cloud as well)

BMurri (approved):

PR #598, opened 2024-02-16
Add global exception handler that logs unhandled controller exceptions

### LINKED ISSUE #599: Unhandled controller exceptions might not be logged
**Describe the bug**
If a controller has an unhandled exception, it might not be logged in the TES logs.

MattMcL4475:
Resolves #599 
- Add global exception handler that logs unhandled controller exceptions
- Simplify the one instance where the application could be stopped, and log the task state

### DIFF
--- a/src/Tes/Repository/PostgreSqlCachingRepository.cs
+++ b/src/Tes/Repository/PostgreSqlCachingRepository.cs
@@ -57,15 +57,15 @@ namespace Tes.Repository
             writerWorkerTask = Task.Run(() => WriterWorkerAsync(writerWorkerCancellationTokenSource.Token))
                 .ContinueWith(async task =>
                 {
-                    switch (task.Status)
+                    if (task.Status == TaskStatus.Faulted)
                     {
-                        case TaskStatus.Faulted:
-                            Logger.LogCritical("Repository WriterWorkerAsync failed unexpectedly with: {ErrorMessage}.", task.Exception.Message);
-                            break;
-                        case TaskStatus.RanToCompletion:
-                            Logger.LogCritical("Repository WriterWorkerAsync unexpectedly completed.");
-                            break;
+                        Logger.LogCritical("Repository WriterWorkerAsync failed unexpectedly with: {ErrorMessage}.", task.Exception.Message);
                     }
+
+
+                    var errorMessage = $"Repository WriterWorkerAsync unexpectedly ended with TaskStatus: {task.Status}";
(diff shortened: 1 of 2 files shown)


src/TesApi.Web/Startup.cs
```diff
@@ -304,6 +306,30 @@ public void Configure(IApplicationBuilder app)
...
                             logger.LogInformation("Configuring for Production environment");
+
+                            s.UseExceptionHandler(a => a.Run(async context =>
+                            {
+                                var exceptionHandlerFeature = context.Features.Get<IExceptionHandlerFeature>();
+
+                                if (exceptionHandlerFeature != null)
+                                {
+                                    var exception = exceptionHandlerFeature.Error;
+                                    logger.LogError(exception, "An unexpected error occurred");
```
BMurri, line 317: Consider adding " while processing an API request." or something along those lines

src/Tes/Repository/PostgreSqlCachingRepository.cs
```diff
@@ -57,17 +57,16 @@ protected PostgreSqlCachingRepository(Microsoft.Extensions.Hosting.IHostApplicat
...
+                        Logger.LogCritical("Repository WriterWorkerAsync failed unexpectedly with: {ErrorMessage}.", task.Exception.Message);
                     }
 
-                    await Task.Delay(50); // Give the logger time to flush.
+                    var errorMessage = $"Critical issue: the repository WriterWorkerAsync unexpectedly ended with TaskStatus: {task.Status}. The TES application will now be stopped.";
+                    Logger.LogCritical(errorMessage);
+                    Console.WriteLine(errorMessage);
+
+                    await Task.Delay(TimeSpan.FromSeconds(5)); // Give the logger time to flush.
                     hostApplicationLifetime?.StopApplication();
```
BMurri, line 70: I like the much better wordsmithing, but:

RanToCompletion is a critical failure, and there's no exception to tell us how we got there. That's something I absolutely want to know about.

Canceled is the "normal" exit path. If we log that at all, it should be Information and we shouldn't include the word "critical anywhere).

I don't care about the exception when Canceled (because it would just be noise) but it's crucial to log the exception when Failed.

So the previous logic was closer to what we need.

Also, I've seen too many systems fail to flush their logs out to the collection system (which are not on host) and I don't think 5 seconds is adequate

BMurri (changes requested):

BMurri (approved):

PR #606, opened 2024-02-16
Explicitly set URL TES listens on

### LINKED ISSUE #607: If ASPNETCORE_URLS is not set, default port has changed
**Describe the bug**
AKS config specifies:
ASPNETCORE_URLS
http://0.0.0.0:80/

Previously, ASP.NET Core would listen on port 80 if this was NOT set.  However, this has changed in .NET 8.  In Terra, this value is not set, so it's a breaking change.

https://learn.microsoft.com/en-us/dotnet/core/compatibility/containers/8.0/aspnet-port

**Steps to Reproduce**
Run TES without setting ASPNETCORE_URLS

**Expected behavior**
TES listens on "http://0.0.0.0:80/" unless explicitly overriden with: ASPNETCORE_URLS

MattMcL4475:
Resolves #607 

Environment Variable Override: Environment variables are read after the in-code configurations. If an ASPNETCORE_URLS environment variable is set, it overrides the URL specified in the .UseUrls() method due to the environment variable having a higher precedence in the configuration system.

### DIFF
--- a/src/TesApi.Web/Program.cs
+++ b/src/TesApi.Web/Program.cs
@@ -32,11 +32,15 @@ namespace TesApi.Web
         /// <returns><see cref="IWebHostBuilder"/></returns>
         public static IWebHostBuilder CreateWebHostBuilder(string[] args)
         {
-            Console.WriteLine($"TES v{Startup.TesVersion} build {System.Reflection.Assembly.GetExecutingAssembly().GetName().Version}");
+            Console.WriteLine($"TES v{Startup.TesVersion}");
 
             Options.ApplicationInsightsOptions applicationInsightsOptions = default;
-            var builder = WebHost.CreateDefaultBuilder<Startup>(args)
-                .UseUrls("http://0.0.0.0:80");
+            var builder = WebHost.CreateDefaultBuilder<Startup>(args);
+
+            if (string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("ASPNETCORE_URLS")))
+            {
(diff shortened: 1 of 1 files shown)


src/TesApi.Web/Program.cs
```diff
@@ -35,11 +35,12 @@ public static IWebHostBuilder CreateWebHostBuilder(string[] args)
             Console.WriteLine($"TES v{Startup.TesVersion} build {System.Reflection.Assembly.GetExecutingAssembly().GetName().Version}");
 
             Options.ApplicationInsightsOptions applicationInsightsOptions = default;
-            var builder = WebHost.CreateDefaultBuilder<Startup>(args);
+            var builder = WebHost.CreateDefaultBuilder<Startup>(args)
+                .UseUrls("http://0.0.0.0:80");
```
BMurri, line 39: All the examples I see never have explicit settings mixed with configuration. The configuration is one step removed from the settings and it's possible that explicit settings will cause configurations to be ignored. I'd feel safer if this was implemented like this:

```
            if (string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("ASPNETCORE_URLS")))
            {
                builder.UseUrls("http://0.0.0.0:80");
            }
```

BMurri: Or, put this in appsettings.json instead of c#

BMurri (changes requested):

BMurri (approved):

PR #610, opened 2024-02-21
Address another class of retriable IO failures in docker image pulls

### LINKED ISSUE #609: Breaking the docker daemon is a retriable failure
**Describe the bug**
In general, environmental failures should be considered transient, until the maximum number of retries has been performed.

**Steps to Reproduce**
1. Using the tes node runner, start an image pull right before the docker daemon decides to quit,
2. Observe an unhandled `System.Net.Http.HttpRequestException` which contains a `System.IO.IOException:` telling use that reading could not be performed due to an `System.Net.Sockets.SocketException` informing us (through a 104 error code) that the peer reset the network connection. This is externally caused, rarely happens, and if enough retries are attempted, is most likely fully recoverable.

**Expected behavior**
Retries were expected. which should have appeared in the logs.

BMurri:
fixes #609

### DIFF
--- a/src/Tes.Runner/Docker/DockerExecutor.cs
+++ b/src/Tes.Runner/Docker/DockerExecutor.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Net;
 using CommonUtilities;
 using CommonUtilities.Options;
 using Docker.DotNet;
@@ -63,7 +64,8 @@ namespace Tes.Runner.Docker
             dockerPullRetryPolicy = new RetryPolicyBuilder(Options.Create(dockerPullRetryPolicyOptions))
                 .PolicyBuilder.OpinionatedRetryPolicy(Polly.Policy
                     .Handle(IsNotAuthFailure)
-                    .Or<IOException>())
+                    .Or<IOException>()
+                    .Or<HttpRequestException>(e => e.InnerException is IOException || e.StatusCode >= HttpStatusCode.InternalServerError || e.StatusCode == HttpStatusCode.RequestTimeout))
                 .WithRetryPolicyOptionsWait()
                 .SetOnRetryBehavior(logger)
                 .AsyncBuild();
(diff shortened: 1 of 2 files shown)


src/Tes.RunnerCLI/Program.cs
```diff
@@ -13,5 +13,13 @@ static async Task<int> StartUpAsync(string[] args)
     CommandFactory.CreateExecutorCommand(rootCommand);
     CommandFactory.CreateDownloadCommand(rootCommand);
 
-    return await rootCommand.InvokeAsync(args);
```
giventocode, line 16: Is this code really needed? The command invocation should already return the error code, and the exception should be thrown and logged. If was added for troubleshooting while running locally, you can change the behavior of the runner to output to the standard out instead of storage by not setting the event log sink in the task definition.

src/Tes.Runner/Docker/DockerExecutor.cs
```diff
@@ -63,7 +64,8 @@ public DockerExecutor(IDockerClient dockerClient, IStreamLogReader streamLogRead
             dockerPullRetryPolicy = new RetryPolicyBuilder(Options.Create(dockerPullRetryPolicyOptions))
                 .PolicyBuilder.OpinionatedRetryPolicy(Polly.Policy
                     .Handle(IsNotAuthFailure)
-                    .Or<IOException>())
+                    .Or<IOException>()
+                    .Or<HttpRequestException>(e => e.InnerException is IOException || e.StatusCode >= HttpStatusCode.InternalServerError || e.StatusCode == HttpStatusCode.RequestTimeout))
```
giventocode, line 68: Make sense to me. It would be nice to have tests to avoid regressions going forward.

src/Tes.Runner/Docker/DockerExecutor.cs
```diff
@@ -177,13 +179,11 @@ private async Task PullImageWithRetriesAsync(string imageName, string? tag, Auth
         {
             logger.LogInformation(@"Pulling image name: {ImageName} image tag: {ImageTag}", imageName, tag);
 
-            await dockerPullRetryPolicy.ExecuteWithRetryAsync(async () =>
-            {
-                await dockerClient.Images.CreateImageAsync(
+            await dockerPullRetryPolicy.ExecuteWithRetryAsync(
```
MattMcL4475, line 182: are we sure this is equivalent to the previous?

BMurri: Other than removing a large compiler-provided state machine, slowing performance for no gain (other than possibly a slightly better exception call stack), this is equivalent to the previous.

src/Tes.RunnerCLI/Program.cs
BMurri: Any exception reaching here was either already logged (before being rethrown) or was never logged (in which case this might give it more visibility in very limited circumstances) but without this change it would not have prevented a zero exit code.

Right now I'm not going to make perfect the impediment of good. If more needs to be done here, we should do it systemically and correctly as a separate issue. Please open one if needed,

src/Tes.Runner/Docker/DockerExecutor.cs
BMurri: #613

PR #615, opened 2024-02-24
Add default scope to batch url

jsaun:
Part of https://github.com/microsoft/CromwellOnAzure/issues/532 
Missed in https://github.com/microsoft/ga4gh-tes/pull/608

### DIFF
--- a/src/TesApi.Web/AzureProxy.cs
+++ b/src/TesApi.Web/AzureProxy.cs
@@ -101,7 +101,7 @@ namespace TesApi.Web
             else
             {
                 location = batchAccountInformation.Region;
-                batchClient = BatchClient.Open(new BatchTokenCredentials(batchAccountInformation.BaseUrl, () => GetAzureAccessTokenAsync(CancellationToken.None, "https://batch.core.windows.net/")));
+                batchClient = BatchClient.Open(new BatchTokenCredentials(batchAccountInformation.BaseUrl, () => GetAzureAccessTokenAsync(CancellationToken.None, "https://batch.core.windows.net//.default")));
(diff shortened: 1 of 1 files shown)


src/TesApi.Web/AzureProxy.cs
```diff
@@ -101,7 +101,7 @@ public AzureProxy(IOptions<BatchAccountOptions> batchAccountOptions, BatchAccoun
             else
             {
                 location = batchAccountInformation.Region;
-                batchClient = BatchClient.Open(new BatchTokenCredentials(batchAccountInformation.BaseUrl, () => GetAzureAccessTokenAsync(CancellationToken.None, "https://batch.core.windows.net/")));
+                batchClient = BatchClient.Open(new BatchTokenCredentials(batchAccountInformation.BaseUrl, () => GetAzureAccessTokenAsync(CancellationToken.None, "https://batch.core.windows.net//.default")));
```
BMurri, line 104: This one should only be one slash.

BMurri (approved):
Thanks for catching this

src/TesApi.Web/AzureProxy.cs
jsaun: Matt pointed out that double slash is the convention for public cloud actually https://github.com/microsoft/CromwellOnAzure/pull/748#discussion_r1495027845

BMurri: That only applies to the public cloud resource manager (control plane). It does not apply to any data-plane system, nor to the resource manager on the sovereign clouds.

PR #620, opened 2024-02-27
Make docker daemon error parsing more resilient

### LINKED ISSUE #618: Docker daemon InternalServerError failures are incorrectly parsed
**Describe the bug**
When parsing the error message returned with `InternalServerError` status responses from the docker daemon, an incorrect assumption was made that the first portion of the three-part message would not contain `:` characters (it was already known that the third part often would). This leads to a lack of attempting to authenticate pulls for some images that require authentication to pull.

**Steps to Reproduce**
Use an image requiring authentication in a task. Depending on the daemon version/state/who-knows-what, you may get an exception associated with a response body where there are one or more colon characters inside embedded quotes in the first part of the message.

**Expected behavior**
All error messages associated with `InternalServerError` statuses are correctly categorized as either an authentication-related issue or not.

**Additional context**
Different daemons, and possibly the same daemons at different times, report the same issues with different details and in different ways. The documentation does NOT describe all of them, and much of the needed information is either found through experimentation or via stackoverflow searches.

BMurri:
fixes #618

### DIFF
--- a/src/Tes.Runner/Docker/DockerExecutor.cs
+++ b/src/Tes.Runner/Docker/DockerExecutor.cs
@@ -28,14 +28,15 @@ namespace Tes.Runner.Docker
         // Exception filter to exclude non-retriable errors from the docker daemon when attempting to pull images.
         private static readonly Func<DockerApiException, bool> IsNotAuthFailure = e =>
             // Immediately fail calls with either 'Unauthorized' or 'Forbidden' status codes
-            !(e.StatusCode == System.Net.HttpStatusCode.Unauthorized || e.StatusCode == System.Net.HttpStatusCode.Forbidden ||
+            !(e.StatusCode == HttpStatusCode.Unauthorized || e.StatusCode == HttpStatusCode.Forbidden ||
                 // Immediately fail calls with a status code of 'InternalServerError' and
                 // an HTTP body consisting of a JSON object with a single string property named "message" where the content is a colon-delimited string of three parts:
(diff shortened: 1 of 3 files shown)


src/Tes.Runner/Docker/DockerExecutor.cs
```diff
@@ -28,14 +28,15 @@ public class DockerExecutor
...
-                // a user-readable description of the failure, the status code from the remote registry server, and the textual description of that status code.
-                // Note that complete validation of the structure of the error report object is not performed. If a one word part after the first colon (and before any
-                // second colon) isn't found, this isn't the failure we are looking for and we'll assume it's retriable.
-                (e.StatusCode == System.Net.HttpStatusCode.InternalServerError &&
-                    new[] { "unauthorized", "forbidden" }.Contains(e.ResponseBody?.Split(':').Skip(1).FirstOrDefault()?.Trim().ToLowerInvariant() ?? string.Empty)));
+                // a user-readable rendition/description of the failure, the status code from the remote registry server, and the textual description of that status code
+                // (often with a web link). Note that complete validation of the structure of the error report object is not performed.
+                (e.StatusCode == HttpStatusCode.InternalServerError && (
+                    (e.ResponseBody?.Contains(": unauthorized:") ?? false) ||
```
MattMcL4475, line 37: seems more future proof to just check if the responsebody contains "unauthorized" or "forbidden" in a case insensitive way

MattMcL4475 (approved):

PR #622, opened 2024-02-27
Update of start-up script to download tes_vm_perf.tgz and execute boo…

jlester-msft:
…tstrapper

### DIFF
--- a/src/TesApi.Web/BatchNodeScriptBuilder.cs
+++ b/src/TesApi.Web/BatchNodeScriptBuilder.cs
@@ -69,8 +69,9 @@ namespace TesApi.Web
         /// </summary>
         /// <param name="runnerBinaryUrl"></param>
         /// <param name="runnerTaskInfoUrl"></param>
+        /// /// <param name="vmPerfArchiveUrl"></param>
         /// <returns></returns>
-        public BatchNodeScriptBuilder WithRunnerFilesDownloadUsingWget(Uri runnerBinaryUrl, Uri runnerTaskInfoUrl)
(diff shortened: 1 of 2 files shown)


src/vm_monitor/vm_monitor_scripts/collect_azure_vm_perf.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
MattMcL4475, line 1: JL to delete

MattMcL4475:
TODO:

1. Add configuration setting; something like `AdvancedVmPerformanceMonitoringEnabled`; default `false`
2.

BMurri:
I'd prefer if we could build the binary download packages in the build rather than check those binaries into git

jlester-msft:
> I'd prefer if we could build the binary download packages in the build rather than check those binaries into git

Agreed this would need some integration into the build pipeline to generate the binary package. 

Right now `src/vm_monitor/build_tes_deployment_archive.sh` contains an example script that will do the building of the telegraf binary + build a .tar.gz of everything. The basic steps are:
1. Build a custom stripped down version of telegraf 
   1. The local build process should probably be skipped in favor of the docker style build in `src/vm_monitor/build_telegraf_remote.sh`. This is setup to build on a remote VM but doesn't need to. It is basically dockerfile build + run the docker to copy out the telegraf binary
2. Make a flattened .tar.gz file with the binary + the scripts needed
3. Put that binary somewhere consumable

PR #643, opened 2024-03-07
Add support to determine calling workflow engine

### LINKED ISSUE #161: Could not parse path '//rc' - misleading log message when non-Cromwell caller
**Describe the bug**
`Could not parse path '//rc'.` error in the logs when a TES task is submitted from a non-Cromwell system

**Steps to Reproduce**
`{"executors":[{"image":"ubuntu:22.04","command":["echo","hello"]}]}`

**Expected behavior**
No log message for this scenario; we can assume it's a non-Cromwell system

BMurri:
required by microsoft/CromwellOnAzure#761


src/TesApi.Web/Controllers/TaskServiceApi.cs
```diff
@@ -160,17 +160,9 @@ public virtual async Task<IActionResult> CreateTaskAsync([FromBody] TesTask tesT
 
             tesTask.State = TesState.QUEUEDEnum;
             tesTask.CreationTime = DateTimeOffset.UtcNow;
+            tesTask.TaskSubmitter = TaskSubmitter.Parse(tesTask);
 
-            // example: /cromwell-executions/test/daf1a044-d741-4db9-8eb5-d6fd0519b1f1/call-hello/execution/script
-            tesTask.WorkflowId = tesTask
```
MattMcL4475, line 165: is this still being set somewhere?

MattMcL4475 (approved):

src/TesApi.Web/Controllers/TaskServiceApi.cs
BMurri: Yes. The `WorkflowId` property now only reads its value from the attached `TaskSubmitter`-derived object, which parses it out from the `TesTask` (currently only for Cromwell-submitted tasks, others are TBD). That way, the same `WorkflowId` property will extend to all workflow engines that implement and provide task metadata on whatever analogous concept exists for those other engines, and it will be `null` for engines that don't provide a means to determine it (or that we don't recognize, yet).

PR #647, opened 2024-03-11
Add basic SDK and scale test

MattMcL4475:


src/Tes.SDK.Tests/TesClientTests.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Text.Json;
+using Moq;
+using Tes.Models;
+
+namespace Tes.SDK.Tests
+{
+    [TestClass]
+    public class TesClientTests
+    {
+        private ITesClient _client;
```
BMurri, line 13: `private ITesClient _client = null!;` will quiet the warnings. ditto for next line

src/Tes/Models/TesTask.cs
```diff
@@ -60,48 +60,48 @@ public TesTask()
         /// </summary>
         /// <value>Input files. Inputs will be downloaded and mounted into the executor container.</value>
         [DataMember(Name = "inputs")]
-        public List<TesInput> Inputs { get; set; }
+        public List<TesInput> Inputs { get; set; } = new List<TesInput>();
```
BMurri, line 63: `public List<TesInput> Inputs { get; set; } = new();` is very concise and does not eliminate any information, and is less distracting when reviewing. ditto for the others in this file.

BMurri: Or, possibly (see if the complier accepts it) `public List<TesInput> Inputs { get; set; } = [];`

src/Tes.SDK/TesClient.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
+                {
+                    _httpClient.Dispose();
```
MattMcL4475, line 167: Quick digression here :)  Does this make sense?  The caller provided the HttpClient, so it doesn't seem like this class' responsibility to dispose of it.  The Microsoft docs say that `HttpClient` is intended to be instantiated once and reused throughout the life of an application.  Now - having said that, I slightly disagree about only one HttpClient, because then one has to be careful that the entire app doesn't modify the default headers.  In this case, I originally use a private static HttpClient, but then I couldn't figure out how to mock it without using putting in the constructor.  Thoughts?

BMurri: I now see the constructor where its passed in, but I don't believe were using it anywhere (except inside of this class) so it's not currently accessible to the caller to dispose.

I'll address that rn, though

BMurri (approved):

src/TesApi.Tests/IntegrationTests.cs
```diff
@@ -0,0 +1,124 @@
...
+                AsyncEnumerable.Repeat(new object(), count).Select(_ => tesClient.CreateTaskAsync(testTask)),
+                new ParallelOptions() { MaxDegreeOfParallelism = 16384 },
+                async (idAsTask, _) => testTaskIds.Add(await idAsTask));
+
+            Console.WriteLine($"Posted {testTaskIds.Count} TES tasks in {sw.Elapsed.TotalSeconds:n3}s");
+            Assert.AreEqual(count, testTaskIds.Count);
+            Assert.AreEqual(count, await tesClient.ListTasksAsync().CountAsync(t => testTaskIds.Contains(t.Id)));
+            sw.Restart();
+
+            while (await tesClient.ListTasksAsync().Where(t => testTaskIds.Contains(t.Id)).AnyAsync(t => t.IsActiveState()))
```
MattMcL4475, line 50: `ListTasksAsync(TesView.Minimal)` would be nice here to save on bandwidths

PR #648, opened 2024-03-15
Update bash scripting, remove jq requirement, add nvme mounting

### LINKED ISSUE #454: Automatically create partition for NVMe disks and use it for images, I/O, and executor Docker volume
**Problem:**
Azure L-series VMs have NVMe drives with no mountpoint.  TES tasks should use these by default if they exist.

**Solution:**
The runner should check if NVMe disks exist, and if so, create a single mountpoint (use RAID0 if more than one disk) in the TES runner.  Use it for Docker image storage, input/output storage, and the main volume for the TES executor.  Consider reformatting every time on task start, so that the previous run is securely deleted (`rm` only removes file references) if more than one task runs on the same machine.  Here is an example of a `Standard_L16S_v3` on startup:

image

### LINKED ISSUE #616: Azure Batch start-up script optimization
**Describe the bug**
The current start-up script optionally installs jq. This install can take anywhere between 10-15seconds and since it is only sometimes installed we can't rely upon jq to always be available.

The current bash script is essentially:
```bash
# /usr/bin/bash -c 'trap "echo Error trapped; exit 0" ERR; sudo touch tmp2.json &&\
(sudo cp /etc/docker/daemon.json tmp1.json || sudo echo {} >tmp1.json) &&
    sudo chmod a+w tmp?.json &&
if fgrep "$(dirname "$(dirname "$AZ_BATCH_NODE_ROOT_DIR")")/docker" tmp1.json; 
    then echo grep "found docker path"; 
elif [ $? -eq 1 ]; 
    then sudo apt-get install -y jq &&
        jq \.\[\"data-root\"\]=\""$(dirname "$(dirname "$AZ_BATCH_NODE_ROOT_DIR")")/docker"\" tmp1.json >>tmp2.json &&
        sudo cp tmp2.json /etc/docker/daemon.json &&
        sudo chmod 644 /etc/docker/daemon.json &&
        sudo systemctl restart docker &&
        echo "updated docker data-root";
else (echo "grep failed" || exit 1); 
fi
```

The change is to replace the jq command:
```bash
jq \.\[\"data-root\"\]=\""$(dirname "$(dirname "$AZ_BATCH_NODE_ROOT_DIR")")/docker"\" tmp1.json >>tmp2.json &&
```

With the following single line of python3:
```python
python -c "import json,os;data=json.load(open(\"tmp1.json\"));data[\"data-root\"]=os.path.join(os.path.dirname(os.path.dirname(os.getenv(\"AZ_BATCH_NODE_ROOT_DIR\"))), \"docker\");json.dump(data, open(\"tmp2.json\", \"w\"), indent=2);"
```
In a more human readable form this python3 script does the same thing the jq command does:
```python
import json,os

# Load the JSON data from tmp1.json
data=json.load(open("tmp1.json"))
data["data-root"]=os.path.join(os.path.dirname(os.path.dirname(os.getenv("AZ_BATCH_NODE_ROOT_DIR"))), "docker")
# Write out the data struct to tmp2.json, with pretty-printing
json.dump(data, open("tmp2.json", "w"), indent=2)
```

This takes the start-up script runtime down to <1s instead of 10-15s. Python3 is available on all Ubuntu 20.04+ versions and CentOS 7.7+, and the script only uses base python3 packages.

**Suggested changes**
Change the start-up to script to this one liner:
```bash
/usr/bin/bash -c 'trap "echo Error trapped; exit 0" ERR; sudo touch tmp2.json && (sudo cp /etc/docker/daemon.json tmp1.json || sudo echo {} > tmp1.json) && sudo chmod a+w tmp?.json && if fgrep "$(dirname "$(dirname "$AZ_BATCH_NODE_ROOT_DIR")")/docker" tmp1.json; then echo grep "found docker path"; elif [ $? -eq 1 ]; then python -c "import json,os;data=json.load(open(\"tmp1.json\"));data[\"data-root\"]=os.path.join(os.path.dirname(os.path.dirname(os.getenv(\"AZ_BATCH_NODE_ROOT_DIR\"))), \"docker\");json.dump(data, open(\"tmp2.json\", \"w\"), indent=2);" && sudo cp tmp2.json /etc/docker/daemon.json && sudo chmod 644 /etc/docker/daemon.json && sudo systemctl restart docker && echo "updated docker data-root"; else (echo "grep failed" || exit 1); fi'
```

jlester-msft:
fixes #616 
fixes #454

This code change makes 3 main changes to `TesApi.Web/BatchScheduler.cs`:
1. Changes the definition of the bash script to use more C# `@`strings. Which makes it much easier to read and work with. Note that `set -e` makes it so any command which fails during bash script execution will trigger an exit. This means some commands like the grep in `nvme_devices=$(nvme list -o json | grep -oP ""\""DevicePath\"" : \""\K[^\""]+"" || true)` needs a `|| true` so that if no match is found the script doesn't end. However, most of the other commands don't require `&&` chaining anymore.
2. Removes the jq install requirement for ubuntu and uses a "here doc" style python3 replacement (see #616)
3. Looks for nvme devices and mounts them as a single RAID0 array onto `${AZ_BATCH_NODE_ROOT_DIR}/tasks/workitems` (`/mnt/batch/tasks/workitems`). Software RAID seems performant and lvm and mdadm RAID0 devices seem to have similar performance. mdadm is used as it is the easier/more mature solution.

These changes have been tested on a few large example workflows. And on some L-series test workloads. Could require additional hardening but seems to be working well enough at the moment.

@BMurri + @MattMcL4475


BMurri:
I recall this being discussed. I don't recall if an issue was opened. I'll go over this (and merge with main, since I'm the last to touch the start-task and I'm certain that's where the merge conflicts mainly are) on Monday.

@jlester-msft could you please respond here with the issue number? Thank you!

jlester-msft:
The L-series NVME issue is #454

BMurri (approved):

BMurri:
@jlester-msft This has been merged with `main` and will definitely be more maintainable. It does substantially alter the 1st main change mentioned in the description, though.

PR #649, opened 2024-03-16
Add option to download runner task file

### LINKED ISSUE #628: [#627 expired SAS tokens] Runner downloads task JSON from URL
**Problem:**
The runner today can both upload AND download files from blob storage by obtaining the needed SAS at the point of need, but it cannot obtain the task JSON that way, which blocks the way to no longer use any SAS tokens in batch task configurations, due to the fact that it requires the task JSON in the local filesystem.

**Solution:**
1. Enable obtaining the authentication metadata (for both TERRA and CoA/standalone deployments) via environment variables.
1. Allow the task JSON to be provided via the command-line as a URL.
1. When the task JSON is a URI, use the current mechanism for downloading files to download the URL and use the content as the task JSON.
1. Consider whether the task JSON should be placed in/near the working directory (in the local filesystem) or if it should be streamed directly to memory and held there until final task completion.

AC
- Unit tests to verify correct behaviors
- Consider integration tests to demonstrate correct behavior.

BMurri:
fixes #628

When the new `url` global option is specified, its argument is downloaded (to the location specified by the `file` option) before the command proceeds. This is accomplished by intercepting all commands and checking for the new option. If found, a subprocess is created to download the json and the `file` option is refreshed. Then the selected command's handler is called. This redirection is facilitated by the use of hidden "shadow" commands. If `url` is not specified the net-net is no observable (from the outside) change in behavior.

Only for the json download, an environment variable is used to obtain the runtime options and transformation selection for the json download. This enables the runner to use its existing code to obtain SAS tokens for the download without any additional arguments or prepositioned files needed. That environment variable is NOT used for any subsequent operations. This will facilitate using this feature in both the start task and in the work tasks without interference.


src/Tes.RunnerCLI/Commands/CommandLauncher.cs
```diff
@@ -99,5 +100,28 @@ public static async Task LaunchesExecutorCommandAsSubProcessAsync(FileInfo file,
 
             HandleResult(results, CommandFactory.ExecutorCommandName);
         }
+
```
BMurri, line 103: note to self: remove this

src/Tes.RunnerCLI/Program.cs
```diff
@@ -17,6 +17,15 @@ static async Task<int> StartUpAsync(string[] args)
     {
         return await rootCommand.InvokeAsync(args);
     }
+    catch (CommandExecutionException ex)
```
BMurri, line 20: Note to self: I think this should go away as well.

BMurri:
Note to reviewers: I did some drive-by cleanup while I was in some of the code files.

src/Tes.RunnerCLI/Commands/NodeTaskUtils.cs
```diff
@@ -31,6 +42,55 @@ public static async Task<NodeTask> DeserializeNodeTaskAsync(string tesNodeTaskFi
...
+        ArgumentNullException.ThrowIfNull(options);
+
+        try
+        {
+            ResolutionPolicyHandler resolutionPolicy = new(options.Value.RuntimeOptions ?? throw new InvalidOperationException($"Environment variable '{nameof(NodeTaskResolverOptions)}' is missing the '{nameof(NodeTaskResolverOptions.RuntimeOptions)}' property."));
+            List<FileInput> sources = [new() { TransformationStrategy = options.Value.TransformationStrategy, SourceUrl = uri.AbsoluteUri, Path = "_" }];  // Path is not used, but it is required
+            var blobUri = (await resolutionPolicy.ApplyResolutionPolicyAsync(sources) ?? []).FirstOrDefault()?.SourceUrl ?? throw new Exception("The JSON data blob could not be resolved.");
+
+            var response = await blobApiHttpUtils.Value.ExecuteHttpRequestAsync(() => new HttpRequestMessage(HttpMethod.Get, blobUri));
+            var nodeTaskText = await response.Content.ReadAsStringAsync();
```
giventocode, line 79: This would not be retried when reading from the stream. We should use the method `ExecuteHttpRequestAndReadBodyResponseAsync`. However, for that one, you need to find out the size of the file so you can create a `PipelineBuffer` to get the data.  This is [how](https://github.com/microsoft/ga4gh-tes/blob/439fd9d5152442d5771a54809b0c223699c344a8/src/Tes.Runner/Transfer/BlobDownloader.cs#L90). This further makes the case for a wrapper around HttpBlobApiUtils, which will facilitate the unit testing of this class as well.

src/Tes.RunnerCLI/Commands/NodeTaskUtils.cs
```diff
@@ -31,6 +42,55 @@ public static async Task<NodeTask> DeserializeNodeTaskAsync(string tesNodeTaskFi
...
+            await File.WriteAllTextAsync(tesNodeTaskFilePath, nodeTaskText);
+        }
+        catch (Exception e)
+        {
+            Logger.LogError(e, "Failed to serialize task JSON file.");
+            throw;
+        }
+    }
+
+    public async Task<NodeTask> ResolveNodeTaskAsync(FileInfo? file, Uri? uri, Lazy<NodeTaskResolverOptions> options)
```
giventocode, line 60: Tests for this functionality are missing, I understand that part of the challenge is that HttpBlobApiUtils is static, but you can create a wrapper that can be mocked very easily.

giventocode (commented):
Lgtm

giventocode (approved):

PR #664, opened 2024-04-05
Fix default serialization

### LINKED ISSUE #659: TES file type enum is wrong
**Describe the bug**
TES file type enum is wrong. Currently `0` or `1`, should be `FILE` and `DIRECTORY`

**Steps to Reproduce**

```
{
  "id": "027dc307384d48d5b0b00c27c71566f7",
  "state": "EXECUTOR_ERROR",
  "name": "CompTest",
  "description": "CompTest",
  "inputs": [
    {
      "name": "access-file",
      "description": "Access file resource via streaming",
      "url": "s3://my-object-store/file1",
      "path": "/data/file1",
      "type": 0,
      "streamable": true
    }
  ],
  "outputs": null,
  "resources": null,
  "executors": [
    {
      "image": "alpine",
      "command": [
        "echo",
        "hello"
      ],
      "workdir": null,
      "stdin": null,
      "stdout": null,
      "stderr": null,
      "env": null
    }
  ],
  "volumes": null,
  "tags": null,
  "logs": [
    {
      "logs": [
        {
          "start_time": "2024-04-05T00:41:44.738487+00:00",
          "end_time": "2024-04-05T00:41:49.014142+00:00",
          "exit_code": 10
        }
      ],
      "metadata": {
        "vm_size": "Standard_D1_v2",
        "vm_family": "standardDv2Family",
        "vm_low_priority": "true",
        "vm_memory_in_gib": "3.5",
        "vm_vcpus_available": "1",
        "vm_regions_available": "[\"australiacentral\",\"australiaeast\",\"australiasoutheast\",\"brazilsouth\",\"brazilsoutheast\",\"canadacentral\",\"canadaeast\",\"centralindia\",\"centralus\",\"centraluseuap\",\"eastasia\",\"eastus\",\"eastus2\",\"eastus2euap\",\"francecentral\",\"germanywestcentral\",\"israelcentral\",\"italynorth\",\"japaneast\",\"japanwest\",\"jioindiawest\",\"koreacentral\",\"koreasouth\",\"northcentralus\",\"northeurope\",\"norwayeast\",\"polandcentral\",\"qatarcentral\",\"southafricanorth\",\"southcentralus\",\"southeastasia\",\"southindia\",\"swedencentral\",\"switzerlandnorth\",\"uaenorth\",\"uksouth\",\"ukwest\",\"westcentralus\",\"westeurope\",\"westindia\",\"westus\",\"westus2\",\"westus3\"]",
        "vm_price_per_hour_usd": "0.0146",
        "vm_hyper_v_generations": "[\"V1\"]",
        "vm_max_data_disk_count": "4",
        "encryption_at_host_supported": "false",
        "vm_resource_disk_size_in_gib": "50.0",
        "failure_reason": "FailureExitCode"
      },
      "start_time": "2024-04-05T00:38:13.7333828+00:00",
      "end_time": "2024-04-05T00:41:49.4501751+00:00",
      "outputs": []
    }
  ],
  "creation_time": "2024-04-05T00:38:10.5109466+00:00"
}
```

**Expected behavior**
 `FILE` and `DIRECTORY`

vsmalladi:
Closes #659


MattMcL4475:
The current change looks like it will just not serialize default properties... but won't resolve #659 

Seems like to resolve #659 something like this is needed:

`opts.SerializerSettings.Converters.Add(new StringEnumConverter());`

BMurri:
> The current change looks like it will just not serialize default properties... but won't resolve #659

The `0` is the default value of the `TesFileType` enumeration, which value does not actually exist (look at the [code](https://github.com/microsoft/ga4gh-tes/blob/43884e2c3495e97804976d4d4125469162f3a0b9/src/Tes/Models/TesFileType.cs#L23)). That zero was written by the `StringEnumConverter`. Adding another one won't do anything (one was already added by the attribute in that code file).

I agree with unit tests, though.

@MattMcL4475

vsmalladi:
@BMurri  should the default be `1` then?

BMurri:
My understanding is that the server is expected to only provide missing types on outputs, per the discussions associated with the removal of the requirement to provide type when submitting the task to the server (a change in 1.1), but maybe I'm recalling that discussion incorrectly. Note also that determination of type when it's missing can only be done with certain kinds of source storage locations, e.g. generic web URLs is one case where directory inputs are not possible to implement.

If the default is changed to 1, then there'll be no way for the server to know it wasn't set in the first place. Making it nullable would be a possibility, but with the proposed fix the entire property will not be written in the JSON when clients retrieve the task when its value is the default (of 0) (along with all other properties that are missing values). Note that arrays that are present but empty will be written with this fix (to prevent writing them would require them to be null).

BMurri:
> I think if it's not provided it should be null.

In C# it's an enumeration. Our choices are:
1) if it's not provided it's completely missing when retrieving the task and its value is `0` when accessing it inside of the server
1) We explicitly make it a `Nullable<>`

`1.` is what the current state of this fix does.

vsmalladi:
> > I think if it's not provided it should be null.
> 
> In C# it's an enumeration. Our choices are:
> 
> 1. if it's not provided it's completely missing when retrieving the task and its value is `0` when accessing it inside of the server
> 2. We explicitly make it a `Nullable<>`
> 
> `1.` is what the current state of this fix does.

So the JSON resturned would be still `0` or `null` from the API

BMurri:
> So the JSON resturned would be still `0` or `null` from the API

In the JSON it is entirely missing. In the POCO model, it is `0`.

vsmalladi:
> In the JSON it is entirely missing. In the POCO model, it is `0`.

That seems reasonable to me.

MattMcL4475 (approved):

PR #673, opened 2024-04-10
Add support for TesInput directory

MattMcL4475:
Adds support for a TES input to be `TesFileType.DIRECTORYEnum`.  The TES server will assume that the `TesInput.Url` is a blob prefix (virtual directory), and will resolve each of the matching blobs as files for the TES runner to download.


src/TesApi.Web/Runner/TaskToNodeTaskConverter.cs
```diff
@@ -424,15 +424,43 @@ private TesInput PrepareLocalFileInput(TesInput input, string defaultStorageAcco
             });
         }
 
-        private static void MapInputs(List<TesInput> inputs, string pathParentDirectory, string containerMountParentDirectory,
+        private async Task MapInputsAsync(List<TesInput> inputs, string pathParentDirectory, string containerMountParentDirectory,
             NodeTaskBuilder builder)
         {
-            inputs?.ForEach(input =>
+            if (inputs == null || inputs.Count == 0)
```
BMurri, line 430: `inputs == null` --> `inputs is null`

Consider: `if ((inputs?.Count ?? 0) == 0)`

src/TesApi.Web/Storage/IStorageAccessProvider.cs
```diff
@@ -92,5 +93,13 @@ public interface IStorageAccessProvider
         /// <param name="blobPath">Path within the reserved blob storage area.</param>
         /// <returns>An Azure Block Blob or Container URL without SAS token.</returns>
         public Uri GetInternalTesBlobUrlWithoutSasToken(string blobPath);
+
+        /// <summary>
+        /// Gets all of the blob URLs for a given blob prefix (virtual directory)
+        /// </summary>
+        /// <param name="blobVirtualDirectory">The URI for the blob prefix (virtual directory)</param>
+        /// <param name="cancellationToken"></param>
```
BMurri, line 101: Copy line 78 to line 101

BMurri:
Looking at the spec and its history/discussions, as well as looking at the nextflow code, `TesFileType` should not assume any default. Instead, if it wasn't set, the input should be checked to determine if it's a "file" or a "directory". Apparently, the server is supposed to update `type` with the result(s) of such a probe right on the `TesInput`s themselves.

vsmalladi:
@BMurri you are correct. We are not updating that right now but could.

vsmalladi:
@MattMcL4475 and @BMurri how can i help push this across?

BMurri:
I plan on picking this up in the morning, but that'll be dependent on the results of an integration test I left running yesterday evening.

vsmalladi:
Awesome let me know if I can help.

BMurri (approved):

PR #681, opened 2024-04-24
Wait for up to 30s to obtain iptables lock

MattMcL4475:
Addresses: https://github.com/microsoft/ga4gh-tes/issues/682


src/Tes.Runner/Docker/NetworkUtility.cs
```diff
@@ -101,6 +102,14 @@ private async Task<(string Output, string Error)> RunIptablesCommandAsync(string
                 case 0:
                     return (output, error);
                 case 4:
+                    if (error?.Contains("Another app is currently holding the xtables lock", StringComparison.OrdinalIgnoreCase) == true)
+                    {
+                        // "Another app is currently holding the xtables lock. Perhaps you want to use the -w option?"
+                        var invalidOperationException = new InvalidOperationException($"xtables is locked by another app. Error: {error}");
```
BMurri, line 108: I'm interested in knowing if the caller to the methods in this class is retrying these calls, and if so, which exceptions result in retries.

The answer isn't needed to merge this code, though

BMurri (approved):

src/Tes.Runner/Docker/NetworkUtility.cs
MattMcL4475: No they are not being retried anywhere

PR #689, opened 2024-04-26
TaskExecutionScriptingManager not thread safe

MattMcL4475:
https://github.com/microsoft/ga4gh-tes/issues/690

### DIFF
--- a/src/TesApi.Tests/Runner/TaskExecutionScriptingManagerTests.cs
+++ b/src/TesApi.Tests/Runner/TaskExecutionScriptingManagerTests.cs
@@ -27,7 +27,6 @@ namespace TesApi.Tests.Runner
         private TaskExecutionScriptingManager taskExecutionScriptingManager;
         private Mock<TaskToNodeTaskConverter> taskToNodeTaskConverterMock;
         private Mock<IStorageAccessProvider> storageAccessProviderMock;
-        private Mock<BatchNodeScriptBuilder> batchNodeScriptBuilderMock;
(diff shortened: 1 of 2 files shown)


src/TesApi.Web/Runner/TaskExecutionScriptingManager.cs
```diff
@@ -126,7 +122,7 @@ private async Task<Uri> CreateAndUploadBatchScriptAsync(TesTask tesTask, Uri nod
         {
             logger.LogInformation($"Creating and uploading Batch script for Task ID: {tesTask.Id}");
 
-            var batchNodeScript = batchNodeScriptBuilder
+            var batchNodeScript = new BatchNodeScriptBuilder()
```
BMurri, line 125: Suggestion: `BatchNodeScriptBuilder batchNodeScript = new()`

src/TesApi.Tests/Runner/TaskExecutionScriptingManagerTests.cs
```diff
@@ -90,9 +87,6 @@ public async Task PrepareBatchScript_ValidTask_CreatesAndUploadsExpectedAssetsTo
                     It.IsAny<NodeTaskConversionOptions>(), It.IsAny<CancellationToken>()),
                 Times.Once);
 
-            //creates the script
```
BMurri, line 93: I'd like to find a way to still validate that the script builder was invoked exactly once. Maybe move the call to create and invoke the builder to its own method and verify that the method is called? However, I wouldn't hold up this PR to restore this part of the test, so if something easy doesn't present itself, don't do more work on it.

BMurri (approved):
lgtm

PR #691, opened 2024-04-26
Update all nuget packages

MattMcL4475:
Update all nuget packages to the latest versions.

### DIFF
--- a/src/Tes/Tes.csproj
+++ b/src/Tes/Tes.csproj
@@ -23,7 +23,7 @@
     <!--Mitigate reported security issues-->
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.2" />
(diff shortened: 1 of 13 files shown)


jsaun (approved):

src/Tes/Tes.csproj
```diff
@@ -12,8 +12,8 @@
 
   <ItemGroup>
     <PackageReference Include="EFCore.NamingConventions" Version="8.0.3" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="8.0.3" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="8.0.2">
+    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="8.0.4" />
```
BMurri, line 15: These will conflict with postgre and they need to stay at 8.0.2 until updates come from postgre

BMurri (changes requested):
The EF libraries need to be reverted and the COA will also need the same library updates as relevant

PR #696, opened 2024-05-01
Simplify TES state to fix ListTasks by state

MattMcL4475:
Addresses: https://github.com/microsoft/ga4gh-tes/issues/698


BMurri (changes requested):
Do the same to `TesFileType` as you did to `TesState`

PR #715, opened 2024-05-28
WX-1594 Support for private ACR access in Terra

jgainerdewar:
I welcome C# style feedback, since I'm new to it.

This change adds two new config inputs: Sam URL and billing profile id. If they're provided, the batch scheduler uses these to fetch a distinct managed identity to use to pull the Docker image for the task. This identity is passed down to the TES Runner. Changes elsewhere will ensure that this identity is assigned to the BatchPool. If it isn't, the log contains a useful error message about the identity not being assigned.


src/Tes.Runner/Docker/ContainerRegistryAuthorizationManager.cs
```diff
@@ -74,12 +79,32 @@ public static bool TryParseAzureContainerRegisteryParts(string imageName, out st
...
+        }
+
+        private async Task<string?> FetchTerraACRActionIdentityAsync(RuntimeOptions runtimeOptions)
+        {
+            if (runtimeOptions.Terra is null || string.IsNullOrWhiteSpace(runtimeOptions.Terra.BillingProfileId) || string.IsNullOrWhiteSpace(runtimeOptions.Terra.SamApiHost))
+            {
+                return null;
+            }
+
+            // TODO test with invalid guid for billing profile id
```
jgainerdewar, line 102: Adding error handling here in case either the Sam request or guid processing fail.

jgainerdewar:
@BMurri Luckily I can see the format check output, but I think you'll need to trigger it to run again with the latest changes, which should have fixed most of the issues (but may have created new ones!).

> The SAM API is called to obtain an azure managed identity that has Acr Pull permissions in one or more azure container registries from which docker images can be retrieved, correct?
Yes, correct. 

> Will that MI ever be different than the one passed in `workflow_execution_identity`?
Yes, when it exists at all it will always be different. The `workflow_execution_identity` is associated with a single user, and the identity we'll use to pull ACR images acts more like a system-level identity. Customers will grant that single identity ACR Pull on their ACRs, and Terra will manage which users in that customer organization are permitted to use that identity to pull images. Happy to meet and talk in more detail about how this works if that helps.

BMurri:
@jgainerdewar I sent you an email

jgainerdewar:
@BMurri Still doing some final testing and planning to add more unit testing, but this is ready for another look.

src/Tes.ApiClients/TerraSamApiClient.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// <summary>
+    /// Terra Sam api client
+    /// Sam manages authorization and IAM functionality
+    /// </summary>
+    public class TerraSamApiClient : TerraApiClient
+    {
+        private const string SamApiSegments = @"/api/azure/v1";
+
+        // TODO configure cache and turn caching of queries back on
+        private static readonly IMemoryCache SharedMemoryCache = new MemoryCache(new MemoryCacheOptions());
```
jgainerdewar, line 21: Ideally we would cache these responses but for no more than a guaranteed short time window (5 minutes?). Does that feel like a reasonable expectation for this caching library? Or should we make a request every time if we want to guarantee a certain level of freshness?

BMurri: A 5-minute cache time seems perfectly reasonable to me.

giventocode (commented):
Looks good to me! Only one point/suggestion to consider.

src/Tes.Runner/Docker/ContainerRegistryAuthorizationManager.cs
```diff
@@ -76,10 +80,20 @@ public static bool TryParseAzureContainerRegisteryParts(string imageName, out st
 
         public ContainerRegistryContentClient CreateContainerRegistryContentClientWithAcquireAuthTokenPolicy(Uri endpoint, string repositoryName, RuntimeOptions runtimeOptions, Action<string> onCapture)
         {
+            // Use the ACR pull identity to generate a token if one is defined, otherwise fall back to the node identity.
```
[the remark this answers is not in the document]
jgainerdewar: Good call, done!

jgainerdewar:
@BMurri This is complete and ready for review, though I'm trying to figure out why, when running with this image, I get tasks failing in Batch with ExitCode 10 any time I have more than one running at once.

BMurri (approved):

PR #717, opened 2024-06-04
Preview: GPU support

BMurri:
This is an initial implementation of GPU support in TES. It has some current limitations:

* assumes compute nodes have outgoing public internet access
* Only NC, ND, and NVv3 SKUs are supported currently supported.
* The `backend_parameters` in the task resources must include `vm_size` set to a specific SKU within the above-described series of VM families.

This follows a combination of NVIDIA, Docker, and Azure documentation for enabling containers to use GPUs. Specifically, this:

* uses the [azure vm extension](https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/hpccompute-gpu-linux) to ensure the CUDA drivers are installed
* installs the [NVIDIA container toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/index.html)
* implements (a portion of) the `--gpus all` [docker cli implementation](https://github.com/docker/cli/pull/1714)

Things in the NVIDIA documentation that are NOT implemented (it's unclear what should or should not be implemented based on our use cases)

* selectable `driver capabilities` mounted inside docker containers
* environment variable advanced configuration (OCI spec)
* support for Container Device Interface (CDI)

Note that the azure documentation for the VM extension for GPU support on linux points the reader to the following EULA: https://www.nvidia.com/en-us/data-center/products/nvidia-ai-enterprise/eula/

addresses microsoft/CromwellOnAzure#356


BMurri:
This has been tested and it works, as per above description.

TODO in order to move from Preview to fully supported option (some of these may be debated):
* Check the V1 and V2 images we point to today, to determine if the drivers, tools and configuration already exist. If so, this can be mostly reversed (with the exception of the flags/runner) and we are done.
* If not, or if partially done, we will need to stage appropriate binaries during deployment and adjust the scripts so that private (and Terra) can be accommodated.

I this this, as a preview, can possibly be shipped this week.

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1094,18 +1147,28 @@ private async Task<BatchModels.StartTask> GetStartTaskAsync(string poolId, Batch
                 cmd.Append($" && {CreateWgetDownloadCommand(await UploadScriptAsync(script, new(await ReadScript(vmFamilyStartupScript))), $"{BatchNodeTaskWorkingDirEnvVar}/{script}", setExecutable: true)} && {BatchNodeTaskWorkingDirEnvVar}/{script}");
             }
 
+            switch (vmFamilySeries)
+            {
+                case VmFamilySeries.standardNFamily:
+                    extensions = extensions.Append(new(name: "gpu", publisher: "Microsoft.HpcCompute", type: "NvidiaGpuDriverLinux", autoUpgradeMinorVersion: true) { TypeHandlerVersion = "1.6" });
```
MattMcL4475, line 1153: should we pull this out as a const for easier future upgrades?

src/Tes.Runner/Docker/DockerExecutor.cs
```diff
@@ -193,22 +194,31 @@ private async Task<MultiplexedStream> StreamStdOutAndErrorAsync(string container
...
+                    HostConfig = new()
                     {
                         AutoRemove = true,
-                        Binds = volumeBindings
+                        Binds = volumeBindings,
+                        DeviceRequests = gpus.GetValueOrDefault()
+                            ? [new()
+                            {
+                                Driver = "nvidia",
+                                Count = -1,
```
MattMcL4475, line 217: should we default to 1?

src/Tes.Runner.Test/Docker/DockerExecutorTests.cs
```diff
@@ -54,7 +54,7 @@ public async Task RunOnContainerAsync_DockerClientReturnsAuthNeeded_CallsContain
             Models.RuntimeOptions runtimeOptions = new();
             try
             {
-                var result = await executor.RunOnContainerAsync(new("msftsc022830.azurecr.io/broadinstitute/gatk", "4.5.0.0-squash", [""], default, default, runtimeOptions));
+                var result = await executor.RunOnContainerAsync(new("msftsc022830.azurecr.io/broadinstitute/gatk", "4.5.0.0-squash", [""], default, default, runtimeOptions, default));
```
MattMcL4475, line 57: should we use a public image instead?

src/TesApi.Web/scripts/config-n-gpu-apt.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/bash
+trap "echo Error trapped; exit 0" ERR
+# set -e will cause any error to exit the script
+set -e
+curl -fsSL https://nvidia.github.io/libnvidia-container/gpgkey | sudo gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg \
```
MattMcL4475, line 5: can you add a comment with a link to the NVIDIA docs if you have one?

MattMcL4475 (approved):

src/Tes.Runner/Docker/DockerExecutor.cs
BMurri: -1 means "all". There's no reason to arbitrarily limit access from GPUs the customer will be paying for.

src/Tes.Runner.Test/Docker/DockerExecutorTests.cs
BMurri: This is a unit test. No call to the docker engine is actually made.

Having said that, placeholder image values would be appropriate here (e.g. "executor/image:tag")

src/TesApi.Web/BatchScheduler.cs
BMurri: It's not at all clear what part(s) of this, if any, would change over time. I added the link to the doc for this specific extension.

src/TesApi.Web/scripts/config-n-gpu-apt.sh
BMurri: Done

giventocode (approved):

src/TesApi.Web/BatchScheduler.cs
```diff
@@ -1067,23 +1117,26 @@ private async Task<BatchModels.StartTask> GetStartTaskAsync(string poolId, Batch
 
             if (!dockerConfigured)
             {
-                var packageInstallScript = machineConfiguration.NodeAgentSkuId switch
+                var packageInstallScript = nodeOs switch
                 {
-                    var s when s.StartsWith("batch.node.ubuntu ", StringComparison.OrdinalIgnoreCase) => "echo \"Ubuntu OS detected\"",
-                    var s when s.StartsWith("batch.node.centos ", StringComparison.OrdinalIgnoreCase) => "sudo yum install epel-release -y && sudo yum update -y && sudo yum install -y wget",
+                    NodeOS.Ubuntu => "echo \"Ubuntu OS detected\"",
+                    NodeOS.Centos => "sudo yum install epel-release -y && sudo yum update -y && sudo yum install -y wget",
```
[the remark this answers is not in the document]
MattMcL4475: agreed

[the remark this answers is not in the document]
BMurri: This functionality is already present, so removing it in a refactor would be an unexpected behavior change.

Having written that, Centos is going away soon, so this will be removed when support for that OS is removed from Azure.

BMurri:
Remaining work: Ensure that this functionality is either turned off or can be safely used if there is no public internet access (I don't know if the VM extensions can be reached from a private virtual network with no public IP access).

Bonus work: if we recognize that the image already has the drivers, just pass them through (don't try to reinstall them).

BMurri:
For the future: the actions taken in the start task for this are not needed for certain VM images, so we can reduce the node startup times (faster tasks, lower costs).

PR #718, opened 2024-06-05
Add convenient initialization of TesClient with a TesCredentials

BMurri:


src/Tes.SDK/TesClient.cs
```diff
@@ -80,6 +80,10 @@ public TesClient(Uri baseUrl, string username, string password)
             ArgumentException.ThrowIfNullOrEmpty(password);
         }
 
+        public TesClient(string scheme, TesCredentials tesCredentials)
```
MattMcL4475, line 83: seems like we should set scheme to "https" by default, or at the very least, have a /// so the developer knows what to put in there

MattMcL4475 (approved):

PR #719, opened 2024-06-10
Address exit code 127 task failures after task's successful completion

BMurri:


MattMcL4475 (approved):
can you test this with both v1 and v2 skus?

Issue #728, opened 2024-06-13
Content-MD5 field is blank for Tes outputs


jsaun:
**Describe the bug**
The TesRunner reworked how the md5 for each output file was calculated and set. The true md5, Content-MD5 field, is now blank which is breaking change for some downstream applications. Add an option to either opt in to older implementation with slower performance or setting the new block md5 into the Content-MD5 field.

PR #729, opened 2024-06-14
[728-1] Runner calculates MD5 checksum for uploads. 

giventocode:
Applies to #728 

In this PR:
 - Added a new property to the node task: `RuntimeOptions.SetContentMd5OnUpload`. When set to `true`, the runner will compute the MD5 for the files before the block list is committed.
 - This change lays the groundwork for implementing hash validation when downloading data using DRS.
 - By default, the setting is false since calculating the MD5 has a performance impact.
 
Pending Work (to complete: #728):
 - Expose this option in the TES server as a runtime configuration, with a default value of `false`.
 - Modify the CoA deployer to include this new configuration option.

### DIFF
--- a/src/Tes.Runner/Transfer/BlobOperationPipeline.cs
+++ b/src/Tes.Runner/Transfer/BlobOperationPipeline.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Security.Cryptography;
 using System.Threading.Channels;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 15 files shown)


src/Tes.Runner/Transfer/BlobOperationPipeline.cs
```diff
@@ -50,10 +51,32 @@ protected BlobOperationPipeline(BlobPipelineOptions pipelineOptions, Channel<byt
...
+
+        using var md5 = MD5.Create();
+
+        await using var stream = new FileStream(filePath, FileMode.Open, FileAccess.Read);
+
+        var hash = await md5.ComputeHashAsync(stream);
+
+        Logger.LogInformation("MD5 hash calculated for file: {filePath}.", filePath);
+
+        return BitConverter.ToString(hash).Replace("-", "").ToLowerInvariant();
```
BMurri, line 77: ditto

BMurri (approved):

PR #732, opened 2024-06-14
Prevent previous docker `pull` from breaking the next task on the compute node

BMurri:


MattMcL4475 (commented):
LGTM

src/Tes.Runner/Docker/DockerExecutor.cs
```diff
@@ -129,7 +129,17 @@ public async Task NodeCleanupAsync(ExecutionOptions executionOptions)
             {
                 if (!IsLastImageSame(ToImageNameWithTag(executionOptions.ImageName, executionOptions.Tag), out var previousImage) && previousImage is not null)
                 {
-                    await DeleteImageAsync(previousImage);
+                    try
+                    {
+                        await DeleteImageAsync(previousImage);
+                    }
+                    catch (DockerApiException ex) when (ex.StatusCode == HttpStatusCode.BadRequest && ex.ResponseBody.Contains("invalid reference format", StringComparison.OrdinalIgnoreCase))
```
giventocode, line 136: What is the purpose of this catch statement?  As it is, we will eat the exception only in one case, all other will be thrown and caught in `RootCommandNodeCleanupAsync`. A clearer approach would be to catch all exceptions here, rename the method to `TryDeleteDockerImagesAsync`. In this method I'd log the exception as well, even if we know it won't be streamed, as it could help with troubleshooting if we are able to get the stdout/sterr from the node somehow.

giventocode (approved):

PR #741, opened 2024-07-15
Handle whitespace in task creation

### LINKED ISSUE #742: Empty/whitespace shall have same effect as null in Tes.Input.URL / Content
**Describe the bug**
If a TES task has " " or "" set for the Content or URL, it could cause a BadRequest, even if it's inconsequential.

**Steps to Reproduce**
Create a TES task and set the input.Content to empty/whitespace and the input.URL to a real value.

**Expected behavior**
It should be inconsequential.

MattMcL4475:
Resolves #742


MattMcL4475:
Leaving as is to allow creation of empty files

PR #762, opened 2024-07-29
Terra: Override allowed-vm-sizes location

### LINKED ISSUE #740: support specifying source for `allowed-vm-sizes`
**Problem:**
When TES is configured to use Terra, TES reads the `allowed-vm-sizes` file from the containing Terra workspace storage container. Any user of the workspace who has permission to run workflows also has permission to write to that storage container; users can modify the file. Thus, the `allowed-vm-sizes` file cannot be used/trusted to restrict the VMs used by this TES instance.

**Solution:**
Terra would like to be able to specify the allowed VM sizes via an environment variable or other deploy-time config which would not be writable by end users. As discussed in person, this could be:
1. a URL from which TES would read `allowed-vm-sizes` instead of - or in addition to - reading from the workspace storage container
2. a delimited list of vm sizes which would take effect instead of - or in addition to - reading the file

We are also open to other solutions if you think an alternative is better.

We have explored Terra's ability to seed this file at TES deploy time, coupled with some kind of monitoring/checksum to ensure the file is not modified. While possible, the ROI on this approach was unattractive.

**Code dependencies**
* CoA, for new and/or existing deployments? _no_
* TES standalone, for new and/or existing deployments? _???_
* Terra, for new and/or existing deployments? _yes; Leo will conditionally specify the allowed vm sizes as it deploys TES_
* Build pipeline? _???_
* Integration tests? _???_

**Additional context**
This feature request is in support of AnVIL Lite.

[Note: Please be sure to set the appropriate label for this issue and tag contributors in the comments to start a discussion]

BMurri:
fixes #740

### DIFF
--- a/src/TesApi.Web/ConfigurationUtils.cs
+++ b/src/TesApi.Web/ConfigurationUtils.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using Tes.Models;
 using TesApi.Web.Management;
 using TesApi.Web.Management.Models.Quotas;
@@ -25,6 +26,7 @@ namespace TesApi.Web
         private readonly IBatchQuotaProvider quotaProvider;
         private readonly IBatchSkuInformationProvider skuInformationProvider;
         private readonly BatchAccountResourceInformation batchAccountResourceInformation;
+        private readonly Uri allowedVmSizesOverride;
 
         /// <summary>
         /// The constructor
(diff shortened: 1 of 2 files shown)


src/TesApi.Web/ConfigurationUtils.cs
```diff
@@ -82,20 +87,20 @@ public async Task<List<string>> ProcessAllowedVmSizesConfigurationFileAsync(Canc
...
             {
-                logger.LogWarning($"Unable to read from {allowedVmSizesUrl.AbsolutePath}. All supported VM sizes will be eligible for Azure Batch task scheduling.");
+                logger.LogWarning("Unable to read from {AllowedVmSizesUrl}. All supported VM sizes will be eligible for Azure Batch task scheduling.", allowedVmSizesUrl.AbsolutePath);
                 return supportedVmSizes.Select(v => v.VmSize).Distinct().ToList();
             }
 
             // Read the allowed-vm-sizes configuration file and remove any previous warnings (those start with "<" following the VM size or family name)
             var allowedVmSizesLines = allowedVmSizesFileContent
-                .Split(new[] { '\r', '\n' })
+                .Split(['\r', '\n'])
```
MattMcL4475, line 103: Going forward, let's agree not to make changes like this unless that line of code is being changed for other reasons

MattMcL4475 (approved):

PR #763, opened 2024-07-29
Pagination and tags filter improvements

### LINKED ISSUE #667: Pagination and tags filter improvements
**Problem:**
Two issues (either one may be considered a bug):
1. It appears that the compliance test cannot interleave `tag_key`' & `tag_value`'s and expects a shorter `tag_value` array to mean that the unmatched `tag_key` elements at the end are the ones intended to be unmatched.
1. Pagination best practices as described [here](https://cloud.google.com/apis/design/design_patterns#list_pagination) include the following advice (`When clients pass in query parameters in addition to a page token, the service must fail the request if the query parameters are not consistent with the page token.`).

**Solution:**
For `1.` we should change the current backup tag filtering method to allow `tagValues.Length` to be `<=` `tagKeys.Length` and to effectively extend `tagValues` to be the same length as `tagKeys` by adding empty string.
For `2.` we should encode all filter arguments into `NextPageToken` (`pageSize` is NOT a filter argument for this purpose) and when finding `page_token` decode the filter arguments. If any arguments are found they must all match, otherwise return `BadRequest`, if none were received use the values decoded from the token.

**Code dependencies**
* `Build pipeline?` No
* `Integration tests?` No

**Additional context**
This is related to the implementation in #342.

BMurri:
fixes #667

### DIFF
--- a/src/TesApi.Web/Controllers/TaskServiceApi.cs
+++ b/src/TesApi.Web/Controllers/TaskServiceApi.cs
@@ -361,10 +361,10 @@ namespace TesApi.Controllers
                     {
                         tags = zippedTags;
                     }
-                    else if (tagKeys.Length == tagValues?.Length)
+                    else if (tagKeys.Length >= tagValues?.Length)
                     {
                         logger.LogWarning("Using backup method to parse tag filters.");
-                        tags = tagKeys.Zip(tagValues, (Key, Value) => (Key, Value))
+                        tags = tagKeys.Zip(tagValues.Concat(Enumerable.Repeat(string.Empty, tagKeys.Length - (tagValues?.Length ?? 0))), (Key, Value) => (Key, Value))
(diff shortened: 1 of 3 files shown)


src/TesApi.Web/Controllers/TaskServiceApi.cs
```diff
@@ -361,10 +361,10 @@ public virtual async Task<IActionResult> ListTasksAsync([FromQuery(Name = "name_
                     {
                         tags = zippedTags;
                     }
-                    else if (tagKeys.Length == tagValues?.Length)
+                    else if (tagKeys.Length >= tagValues?.Length)
```
MattMcL4475, line 364: can we add a unit test for this?

BMurri: Done

MattMcL4475 (approved):

PR #751, opened 2024-07-22
WX-1705 Support DRS Compact Identifiers

jgainerdewar:
This PR adds support for DRS compact identifiers, which are not valid IETL URIs. In order to allow the DRS URI transformation strategy to operate on them without requiring update of all transformation strategies, error handling has been added to `CombinedTransformationStrategy`.

This is in draft because I'll working through some errors caused by recent TES work.

PR #766, opened 2024-08-08
WX-1705 Support DRS Compact Identifiers

BMurri:
This PR replaces #751 by adding a `UriParser` for the DRS schema (supporting both styles). DRS compact identifiers URIs are not valid IETL URIs, which caused failures when creating Uri objects.

Most of this pull request is from @jgainerdewar whom I most graciously thank.


src/CommonUtilities/DrsUriParser.cs
```diff
@@ -0,0 +1,235 @@
...
+                return !path.TrimStart('/').Any(c => c == '/') && !path.TrimStart(':').Any(c => c == ':') && IsAccessionValid(uri.Segments.Last().TrimStart(':'));
+            }
+        }
+
+        private static bool IsCompactIdUri(Uri uri)
+            => uri.OriginalString.StartsWith(_prefix, StringComparison.OrdinalIgnoreCase) && _drsCompactId.IsMatch(uri.OriginalString.AsSpan(_prefixLength));
+
+        // https://ga4gh.github.io/data-repository-service-schemas/docs/#tag/DRS-API-Principles/DRS-IDs
+        // https://ga4gh.github.io/data-repository-service-schemas/docs/more-background-on-compact-identifiers.html#tag/Background-on-Compact-Identifier-Based-URIs
+        [GeneratedRegex("\\A(?<provider_code>[\\._a-z]+?/)?(?<namespace>[\\._a-z]+?):(?<accession>[%-\\.0-9A-Z_a-z~]+?)\\Z", RegexOptions.ExplicitCapture | RegexOptions.CultureInvariant)]
```
MattMcL4475, line 232: this looks pretty gnarly and potentially buggy... do you think we have enough test cases for it?

MattMcL4475 (approved):

src/CommonUtilities/DrsUriParser.cs
```diff
@@ -0,0 +1,235 @@
...
+
+                    if (IsComponent(components, UriComponents.Host))
+                    {
+                        if (builder.Length > 0)
+                        {
+                            builder.Append(Uri.SchemeDelimiter);
+                        }
+
+                        // "provider_code" includes the separating '/' if a provider code was found due to the regex.
+                        builder.Append(match.Groups["provider_code"].Value + match.Groups["namespace"].Value);
```
MattMcL4475, line 105: Any concerns about a null reference when no match groups found?

src/CommonUtilities/DrsUriParser.cs
BMurri: I have test cases to cover with and without `provider_code`, both passing and failing the allowed character set, and other missing elements. This regex only covers Compact ID DRS URI (both styles).

The regex does not cover Hostname DRS URI, which are covered via a combination of [this code utilizing an `https` Uri](https://github.com/microsoft/ga4gh-tes/blob/79581005b8a52e4eda378a4723b1afcabee50180/src/CommonUtilities/DrsUriParser.cs#L123) and [this code limiting the `Path` portion of the Uri](https://github.com/microsoft/ga4gh-tes/blob/79581005b8a52e4eda378a4723b1afcabee50180/src/CommonUtilities/DrsUriParser.cs#L218) but I also have test cases covering both passing and failing (in similar ways) for those as well.

PR #773, opened 2024-08-20
Add parent task ID to execute entire compute graph

MattMcL4475:


BMurri:
@MattMcL4475 You would still have to have the parent task id, but Azure Batch can delay tasks until their parent tasks run, so there's no need to wait for task completion to schedule the next task (bonus, less new code)

PR #807, opened 2024-10-21
Add bwa example.

vsmalladi:


BMurri (approved):
Looks correct insofar as it matches the other examples

vsmalladi:
Whats the best way to have a user run these?

PR #816, opened 2024-11-07
Successfully don't add SAS token on blob storage URLs with no permissions

BMurri:
This is due to `datasettestinputs` no longer accepting SAS tokens.

### DIFF
--- a/src/Tes.Runner.Test/Storage/ArmUrlTransformationStrategyTests.cs
+++ b/src/Tes.Runner.Test/Storage/ArmUrlTransformationStrategyTests.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System.Net;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
 using Azure.Storage.Sas;
--- a/src/Tes.Runner.Test/Storage/ArmUrlTransformationStrategyTests.cs
+++ b/src/Tes.Runner.Test/Storage/ArmUrlTransformationStrategyTests.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Net;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
 using Azure.Storage.Sas;
 using CommonUtilities;
-using CommonUtilities.Options;
 using Moq;
 using Tes.Runner.Models;
 using Tes.Runner.Storage;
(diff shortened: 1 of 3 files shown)


giventocode (commented):
I understand the approach—assuming the endpoint is public if a read-only key is requested but cannot be granted. However, do you think it would be better to simply call the URL with a HEAD request, so no content is returned and use that as the validation gate?

BMurri:
> ...do you think it would be better to simply call the URL with a HEAD request, so no content is returned and use that as the validation gate?

Probably, but that's a bigger PR that I would prefer more bake time for (ArmUrlTransformationStrategy is created separately for each input and output, and given the warnings I'd prefer not to proliferate possibly hundreds of HttpClient instances that are each used maybe once). Would be a very good follow on PR.

giventocode:
> > ...do you think it would be better to simply call the URL with a HEAD request, so no content is returned and use that as the validation gate?
> 
> Probably, but that's a bigger PR that I would prefer more bake time for (ArmUrlTransformationStrategy is created separately for each input and output, and given the warnings I'd prefer not to proliferate possibly hundreds of HttpClient instances that are each used maybe once). Would be a very good follow on PR.

I don't see how it would be a larger PR; the implementation would be simple. You could just create a new method, IsStorageEndPointPublic, that makes an HTTP HEAD call using [this method](https://github.com/microsoft/ga4gh-tes/blob/ee48a9bff8b4a2911df4637999f18e4e4d889abc/src/Tes.Runner/Transfer/BlobApiHttpUtils.cs#L167). If the call returns false, attempt to get the user-delegated key as the code was doing it before. This also addresses the concern of multiple HTTP clients, as the BlobApiHttpUtils use a static/read-only shared client. Additionally, keep in mind that obtaining the user-delegated key is an HTTP operation, so the I/O penalty is already incurred.

giventocode:
> > ...do you think it would be better to simply call the URL with a HEAD request, so no content is returned and use that as the validation gate?
> 
> Probably, but that's a bigger PR that I would prefer more bake time for (ArmUrlTransformationStrategy is created separately for each input and output, and given the warnings I'd prefer not to proliferate possibly hundreds of HttpClient instances that are each used maybe once). Would be a very good follow on PR.

I don't see how it would be a larger PR; the implementation would be simple. You could just create a new method, IsStorageEndPointPublic, that makes an HTTP HEAD call using [this method](https://github.com/microsoft/ga4gh-tes/blob/ee48a9bff8b4a2911df4637999f18e4e4d889abc/src/Tes.Runner/Transfer/BlobApiHttpUtils.cs#L167). If the call returns false, attempt to get the user-delegated key as the code was doing it before. This also addresses the concern of multiple HTTP clients, as the BlobApiHttpUtils use a static shared client, which is considered best practice. Additionally, keep in mind that obtaining the user-delegated key is an HTTP operation, so the I/O penalty is already incurred."

giventocode (approved):
LGTM

PR #822, opened 2025-01-17
Add test node debugging aid

BMurri:
fixes: #823

This is currently implemented as an opt-in. For the `tes` workflow in the k8s cluster, add the `BatchNodes__DebugDelayInMinutes` environment variable with a value that is a decimal number of minutes to delay the task ending when the task runner is returning a non-zero exit code.

Note that updating the deployment with the deployer may remove this variable.

### DIFF
--- a/src/CommonUtilities/AzureServicesConnectionStringCredential.cs
+++ b/src/CommonUtilities/AzureServicesConnectionStringCredential.cs
@@ -71,7 +71,7 @@ namespace CommonUtilities
 
         private void SetInitialState(AzureCloudConfig armEndpoints)
         {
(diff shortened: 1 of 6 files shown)


src/CommonUtilities/AzureServicesConnectionStringCredential.cs
BMurri, line 1: This was changed to fix a CodeQL build break

src/TesApi.Web/Storage/DefaultStorageAccessProvider.cs
BMurri, line 1: This was changed to fix a CodeQL build break

MattMcL4475 (approved):

